PR #3, opened 2025-09-16
Fix revision functionality

ingeniumed:
### Description

The purpose of this PR is to fix the revision functionality. This is a companion PR to [this Gutenberg PR](https://github.com/Automattic/gutenberg/pull/24).

It stores the last revision ID in the post meta, alongside the CRDT Doc details.

This also disables the _edit_lock meta that's used for locking revision functionality to a single user. It doesn't work anywhere else.

### ToDo

- The title doesn't get set correctly sometimes. I haven't been able to resolve this. It seems like the other remote clients are overriding it and my fix used for the blocks doesn't work here as this is a string.
- Verify that the way I'm updating the `vip_rtc_state` is correct, after a revision restore is detected. For us, a restore means the doc is updated so all of that would need to be updated. However, when the next save happens all of this would be re-verified anyways.
- Code cleanup and testing again to make sure I didn't miss anything.

### DIFF
--- a/inc/Compatibility/Compatibility.php
+++ b/inc/Compatibility/Compatibility.php
@@ -72,8 +72,7 @@ final class Compatibility {
 	/**
 	 * Check if the Gutenberg plugin is active.
 	 *
-	 * TODO: Check GUTENBERG_VERSION in production to ensure it is running a
-	 * compatible version.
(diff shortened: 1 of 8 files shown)


ingeniumed:
Closing this off as it's been resolved now in another way.

PR #4, opened 2025-09-16
CompatibilityTest.php: Fix failing test

acicovic:
While working on something else, I noticed that our compatibility class integration test was failing with the latest changes.

This PR fixes the issue, and additionally runs the test with the `VIP_RTC_WS_URL` constant set to "".


tests/Integration/CompatibilityTest.php
```diff
@@ -24,17 +24,21 @@ public function test_should_plugin_load(): void {
...
 		// Activate Gutenberg.
 		activate_plugin( 'gutenberg/gutenberg.php' );
 		self::assertTrue( $is_gutenberg_plugin_active->invoke( null ), 'is_gutenberg_plugin_active should be true' );
-		self::assertTrue( $should_plugin_load->invoke( null ), 'should_plugin_load() should be true' );
+
+		if ( true === $is_websocket_url_defined->invoke( null ) ) {
+			self::assertTrue( $should_plugin_load->invoke( null ), 'should_plugin_load() should be true' );
+		} else {
+			self::assertFalse( $should_plugin_load->invoke( null ), 'should_plugin_load() should be false' );
+		}
```
chriszarate: O_o

We should either require that the constant be defined or mock the constant definition and test both cases.

tests/Integration/bootstrap.php
```diff
@@ -30,6 +30,16 @@
...
 require_once __DIR__ . '/../../vendor/yoast/wp-test-utils/src/WPIntegration/bootstrap-functions.php';
 
+// Define required constants for testing.
+$test_ws_url = getenv( 'TEST_VIP_RTC_WS_URL' );
+if ( false !== $test_ws_url ) {
+	// If the environment variable is set (even if empty), use its value.
+	define( 'VIP_RTC_WS_URL', $test_ws_url );
+} else {
+	define( 'VIP_RTC_WS_URL', 'ws://localhost:1234' );
+}
```
chriszarate: What is the point of this code? Are there tests that depend on a varying definition of `VIP_RTC_WS_URL`? Why not hardcode a test value here, similar to `VIP_RTC_WS_AUTH_SECRET` below?

composer.json
```diff
@@ -16,7 +16,7 @@
     "phpcs-fix": "phpcbf",
     "prepare-ci": "bash bin/install-wp-tests wordpress_test root root localhost",
     "psalm": "psalm.phar --no-cache",
-    "testwp": "@php ./vendor/bin/phpunit --no-coverage --order-by=random -c phpunit-integration.xml"
+    "testwp": "TEST_VIP_RTC_WS_URL=\"\" phpunit --no-coverage -c phpunit-integration.xml --filter=CompatibilityTest && phpunit --no-coverage --order-by=random -c phpunit-integration.xml"
```
chriszarate: 1. Can we split pure test runs and coverage runs into separate commands?
2. ~Using `./vendor/bin/phpunit` is preferable so that we use the version installed by Composer, and don't assume it is installed globally.~

chriszarate: Oh this is composer.json, you can ignore point 2. Sorry, thought this was package.json

inc/Auth/WebSocketAuth.php
```diff
@@ -48,6 +48,7 @@ public static function generate_token(
 
 		// Get the JWT secret from constant
 		if ( defined( 'VIP_RTC_WS_AUTH_SECRET' ) ) {
+			/** @psalm-suppress RedundantCast */
```
chriszarate, line 51: Can we just remove the cast then?

composer.json
acicovic: There are no coverage commands here. Am I misunderstanding the question?

inc/Auth/WebSocketAuth.php
acicovic: If there's a slight chance of ever having this passed as a non-string for any reason, I think that having the cast is good.

I think that Psalm started complaining about this with the addition of the [new code in bootstrap.php](https://github.com/Automattic/vip-real-time-collaboration/pull/4#discussion_r2355642076).

tests/Integration/bootstrap.php
acicovic: This code allows us to test in CompatibilityTest, with `VIP_RTC_WS_URL` being `''` or `'ws://localhost:1234'`. This will produce different results, as the former will make `should_plugin_load()` return `false` and the later `true`.

1. If we think that `VIP_RTC_WS_URL` cannot be "" under any circumstance, we can remove this.
2. If we don't like the current implementation, we can look into a better way to do this. What I was after is having a real constant value instead of trying workarounds.

tests/Integration/CompatibilityTest.php
acicovic: I'm not clear on what is being proposed here. The constant can be `undefined`,  `''` or a non-empty string value. In the first two cases `is_websocket_url_defined()` returns `false`.

tests/Integration/bootstrap.php
chriszarate: Understood, but if we want to test that logical fork, we don't want to run the test suite multiple times with different env values. We should look into mocking the constant ([example](https://docs.mockery.io/en/latest/cookbook/class_constants.html)). But for this PR, I think it's fine to assume that it's defined.

composer.json
chriszarate: No, just misinterpreted it. As noted above, we don't want to do multiple test suite runs. So resolving the other issue will resolve this one. Sorry for the noise.

acicovic:
@chriszarate, I've tried to address all your feedback in https://github.com/Automattic/vip-real-time-collaboration/pull/4/commits/a8fef23055fa6085644d6f5724f6a64edeeb36b6.

Also, I can confirm that Psalm complains about the redundant cast due to the WS constants being set in `bootstrap.php`. We could probably ignore that file if we want to remove the ignore rule.

chriszarate (approved):

PR #6, opened 2025-09-16
Removing post editor lock dialogue

pkevan:
Migrated from: https://github.a8c.com/Automattic/vip-real-time-collaboration/pull/102

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,70 @@
+name: Static analysis
+on:
+  pull_request:
+  push:
+    branches:
+      - trunk
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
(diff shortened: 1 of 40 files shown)


pkevan:
Hmm, so 93c8d2e0729f6a128f4e7de5be48f54124f9aa77 actually breaks the default behaviour of unlocking a post on existing the editor via https://github.com/Automattic/gutenberg/blob/47432d59a650b62836d1eaed990cfd5d7c7bb989/packages/editor/src/components/post-locked-modal/index.js#L138. Not quite sure what is going wrong, but noting for now.

cc @chriszarate

pkevan:
Actually, there is another thing - the call to `wp_ajax_wp_remove_post_lock` which you would expect to remove the lock, but instead it just added more time to the post lock 😂

pkevan:
See: https://core.trac.wordpress.org/ticket/63993

PR #9, opened 2025-09-16
Add GitHub codeowners and dependabot config

shekharnwagh:
Mostly taken from https://github.com/Automattic/remote-data-blocks with additional dependabot configs for `websocket-server` and `yjs-inspector` which also live in this repo.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,8 @@
 # See https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+# This team will be requested for review when someone opens a pull request.
 * @Automattic/vip-cafe
+
+# Don't assign owners for the following files:
+package.json
+package-lock.json
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# See https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+* @Automattic/vip-cafe
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=https://json.schemastore.org/dependabot-2.0.json
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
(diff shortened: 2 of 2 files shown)


acicovic (commented):
If desired and regarding CODEOWNERS, we could ignore changes to `package.json` and `package-lock.json`. Se [here](https://github.com/Parsely/wp-parsely/pull/1286) for an example.

shekharnwagh:
> If desired and regarding CODEOWNERS, we could ignore changes to `package.json` and `package-lock.json`. Se [here](https://github.com/Parsely/wp-parsely/pull/1286) for an example.

Wouldn't we want that someone from the team takes a look at any changes in package.json and package-lock.json , especially to account for things like supply chain attacks ?

acicovic:
> Wouldn't we want that someone from the team takes a look at any changes in package.json and package-lock.json , especially to account for things like supply chain attacks ?

Dependency updates come from dependabot or contributors (in or outside of the team). All these should come as PRs (we shouldn't accept direct pushes to `trunk` or other branches). When reviewing a PR that's either automated or user-submitted, we should audit the changes in our package files. We should be especially suspicious of user-submitted PRs that change dependency if these come from outside of the team.

If all this is in place, it blocks potentially unwanted changes to our package files. However, if we think that leaving as is is more secure or gives us more peace of mind, I have of course no objections.

acicovic (approved):

shekharnwagh:
> > Wouldn't we want that someone from the team takes a look at any changes in package.json and package-lock.json , especially to account for things like supply chain attacks ?
> 
> Dependency updates come from dependabot or contributors (in or outside of the team). All these should come as PRs (we shouldn't accept direct pushes to `trunk` or other branches). When reviewing a PR that's either automated or user-submitted, we should audit the changes in our package files. We should be especially suspicious of user-submitted PRs that change dependency if these come from outside of the team.
> 
> If all this is in place, it blocks potentially unwanted changes to our package files. However, if we think that leaving as is is more secure or gives us more peace of mind, I have of course no objections.

Seems fair enough, added the changes from PR you referenced. We can always revisit this in the future.

acicovic:
Mentioning @chriszarate and @pkevan in case they think we shouldn't have these exceptions.

PR #27, opened 2025-09-17
Add dedicated integration tests workflow

acicovic:
Adding a dedicated integration tests workflow to see how things are behaving across different PRs.


acicovic:
For posterity, the workflow fails against `trunk` (which was my experience), but interestingly only for PHP 8.2. The work in #4 is meant to address the issue, and the tests pass there.

PR #29, opened 2025-09-17
Add GPL 2.0 License

shekharnwagh:
Copied from https://github.com/Automattic/remote-data-blocks


chriszarate (changes requested):
Internal review

PR #30, opened 2025-09-17
Fix CRDT doc invalidation and drop version check

chriszarate:
Persisted CRDT invalidation was broken due to some bad assumptions about "record" vs "object data." The primary issue is the hash was being computed with a record that included `blocks`. Since the blocks are parsed on each initial page load (via `SyncConfig#getInitialObjectData`), they will always have different client IDs and the hashes will mismatch.

We need some custom logic to drop blocks and use `content` instead. As noted in the comments, this should more ideally be controlled by the entity's `SyncConfig`, which could provide the hash implementation directly. However, for now, we want to keep the CRDT persistence implementation in this plugin and outside Gutenberg, along with our choice of hashing mechanisms. 

Minor: We were also not sorting the keys before passing to `JSON.stringify` which might also have led to bad invalidation in edge cases.

We are also dropping the CRDT version check, since that is better implemented by the upstream `SyncProvider`, probably with assistance from the `SyncConfig`. Action must be taken to migrate between versions, and this plugin will not have enough information to conduct that migration. 

In an ideal world, we would have generic types that would help us distinguish between a "raw record" and syncable "object data." For now, we will rename the variables as a primitive signal to help avoid these bad assumptions in the future.

### DIFF
--- a/inc/Editor/CrdtPersistence.php
+++ b/inc/Editor/CrdtPersistence.php
@@ -4,7 +4,6 @@ namespace VIPRealTimeCollaboration\Editor;
 
 use VIPRealTimeCollaboration\Compatibility\Compatibility;
 use function add_action;
-use function post_type_supports;
 use function register_meta;
 
 defined( 'ABSPATH' ) || exit();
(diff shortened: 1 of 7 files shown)


ingeniumed (approved):
Noting that right now, in order to get the new content to show up I need to:

- Either wait for the clients to disconnect and open afresh, OR
- Go to the revisions page on one of the clients and then load the post again

Then the post is updated. Both of those are better than how it was before where it was broken.

PR #36, opened 2025-09-18
Enable restore revisions for all users

ingeniumed:
### Description

This will enable the `restore revisions` workflow for all users by targeting the `_edit_lock` meta. 

To ensure that the blast radius of this filter is minimal, we give back false in only the following situations:

- Revision.php has to be the page that is currently being worked on
- `_edit_lock` has to be the meta being requested
- The post that's been globally stored, should support collaboration and should match the ID for the meta that's being requested

In any other case, the default value is provided back. 

So for cases like third party plugins or code that need that meta, they would hopefully not be impacted because the `pagenow` would not be set to revision.php.

I've also split off the e2e tests from the integration tests workflow, and cleaned it up a bit.

### Follow ups

We should do more testing once this is merged in to see what's reported for third party code.


inc/Overrides/Overrides.php
```diff
@@ -19,5 +22,37 @@ final class Overrides {
...
+		// Skip if not on the revisions.php page, or the meta_key is not _edit_lock.
+		if ( 'revision.php' !== $pagenow || '_edit_lock' !== $meta_key ) {
+			return $value;
+		}
+
+		$supported_post_types = Compatibility::get_supported_post_types();
+
+		// If there is a post and it matches the object ID and is a supported post type, return false to disable the lock.
+		/** @var WP_Post|null $post */
+		if ( $post && $post->ID === $object_id && in_array( $post->post_type, $supported_post_types, true ) ) {
```
chriszarate: When does the `$object_id` not match the post ID?

ingeniumed: It likely shouldn't happen, but I left it in there to be extra safe.

pkevan: it's probably better to grab the post object via `$object_id` than rely on the global `$post` being accurate.

inc/Overrides/Overrides.php
```diff
@@ -19,5 +22,37 @@ final class Overrides {
...
+			return false;
```
chriszarate: This is a nitpick, but we should match the return structure of `get_post_meta`, which [returns an empty string](https://developer.wordpress.org/reference/functions/get_metadata/) if a single meta does not exist.
```suggestion
			return '';
```

inc/Overrides/Overrides.php
```diff
@@ -19,5 +22,37 @@ final class Overrides {
...
+		global $post, $pagenow;
+
+		// Skip if not on the revisions.php page, or the meta_key is not _edit_lock.
+		if ( 'revision.php' !== $pagenow || '_edit_lock' !== $meta_key ) {
+			return $value;
+		}
+
+		$supported_post_types = Compatibility::get_supported_post_types();
+
+		// If there is a post and it matches the object ID and is a supported post type, return false to disable the lock.
```
chriszarate:
```suggestion
		// If there is a post and it matches the object ID and is a supported post type, return an empty string to disable the lock.
```

inc/Overrides/Overrides.php
ingeniumed: I like that, switched to it instead.

inc/Overrides/Overrides.php
ingeniumed: Done

inc/Overrides/Overrides.php
ingeniumed: Done

pkevan (approved):

chriszarate (changes requested):
Not sure why GitHub isn't showing it, but there is a conflict with the GitHub workflows included in this PR

chriszarate (approved):
I fixed the conflict. Feel free to apply those changes in a new PR!

PR #37, opened 2025-09-19
Combine highlighting and cursor into one setting

ingeniumed:
### Description

The goal of this is to combine the `Enable cursors` and `Enable highlighting` into one setting - `Enable cursors` to power both the cursor and block highlighting feature. This was one of the issues we talked about in the last sync, but wasn't converted to a linear issue.

I've kept the option as `Enable cursors` internally and externally. I figured its similar to the Google Docs point which is that it's the same thing - cursors everywhere.

Happy to change this up as needed though.

CleanShot 2025-09-19 at 14 29 36

https://github.com/user-attachments/assets/087e1a96-3b1a-457c-96ec-5454519bce21


alecgeatches (approved):
Looks good to me! Thank you for simplifying the sidebar and code.

PR #40, opened 2025-09-22
Improve integration tests workflow

acicovic:
With this PR, we're improving our integration tests workflow. Notably:

- We include PHP 8.1 testing
- We regroup/rename some integration test commands for clarity and order, also adding script descriptions
- We introduce experimental mode which doesn't fail on deprecation errors
- We run integration tests in code coverage mode under PHP 8.4 using `pcov`
- We include PHP 8.5 testing, against WP trunk and in experimental mode

PR #41, opened 2025-09-22
Update end-to-end tests workflow

acicovic:
## Description

Updates the workflow introduced by @ingeniumed to only perform end-to-end tests. Integration tests are performed by the workflow introduced in #27 and updated in #40.


acicovic:
Noting that we'll also need to do some cleanup of our actions after this gets merged.

image

chriszarate (approved):

PR #43, opened 2025-09-23
Clean up channel state within the websocket server

ingeniumed:
### Description

Currently, the websocket server depends on the `setupWSConnection` implementation from [y-websocket-server](https://github.com/yjs/y-websocket-server/blob/v0.1/src/utils.js#L250). In this implementation, the in-memory map of Yjs docs is only cleaned up on all the connections closing if persistence has been configured. If persistence isn't configured, then those docs are held indefinitely in memory. The only thing that would eject them from memory is if the websocket server runs out of memory or gets restarted.

What I've done is ported over the two files that we use in that entire package, and modified it to eliminate persistence entirely. You can compare the implementation [here](https://github.com/yjs/y-websocket-server/blob/v0.1/src/utils.js) with ours to see the difference. 

The key one being [here](https://github.com/yjs/y-websocket-server/blob/v0.1/src/utils.js#L216) which is to remove the persistence check when there are no connections available within the doc.

### Why is this needed?

This will cause the memory consumption of the websocket server to steadily go up and up. Also, we shouldn't really have the doc remain in memory if there is no one connected to it and there no updates needed to be sent out.

### Follow Up

- Should a timer be added from the moment the last connection is dropped before the doc is cleared from the map? I didn't add this in right now, I kept it so that doesn't happen. It clears it out the moment every connection is closed.
- The code needs to be converted to typescript, and cleaned/improved.

Both of these can be done in follow up PRs.


ingeniumed:
Closes VIP-1843

chriszarate:
Interesting! Curious if you tried using:

1. the included file system persistence (LevelDB)? We could point it to `/tmp`.
1. the `setPersistence` function, which allows you to set a custom persistence (which could be a no-op)?

ingeniumed:
After our discussion, what I've done is:

- Implemented a No-op persistence provider that only implements `bindState` and `writeState`. There is an opportunity to implement debugging statements during development or in the future. For now, nothing is in place.
- Set this No-op persistence provider as the persistence variable, so that the in-memory instance is destroyed and deleted when all the connections are closed. This way, nothing will be held in state indefinitely.

I have tested this and ensured that everything is working as expected.

What I haven't done:

- Implemented a delay after all the connections are closed in removing the doc from in-memory. We can see after this goes in if we want this, or is the existing WP setup that we have using `post_meta` sufficient.

websocket-server/index.ts
```diff
@@ -189,6 +190,13 @@ const server = http.createServer( async ( request, response ) => {
 
 const metricsServer = createMetricsServer();
 
+const noopPersistence = new NoopPersistenceProvider();
+setPersistence( {
+	bindState: ( docName, yDoc ) => noopPersistence.bindState( docName, yDoc ),
+	writeState: ( docName, yDoc ) => noopPersistence.writeState( docName, yDoc ),
+	provider: noopPersistence,
+} );
```
chriszarate:
```suggestion
setPersistence( new NoopPersistenceProvider() );
```

websocket-server/noop-persistence-provider.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * WordPress dependencies
+ */
+import type { CRDTDoc } from '@wordpress/sync';
```
chriszarate: yjs is already a subdependency, so we should install it explicitly and prefer the "real" type over the alias

```suggestion
import type * as Y from 'yjs';
```

websocket-server/noop-persistence-provider.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A no-operation persistence provider for Y.js collaborative documents.
+ *
+ * This provider implements the persistence interface but doesn't actually
+ * persist any data to storage. It is mainly to ensure that, the websocket-server
+ * ejects documents from memory when there are no active connections.
```
chriszarate:
```suggestion
 * persist any data to storage. It is mainly to ensure that the websocket-server
 * evicts documents from memory when there are no active connections.
```

websocket-server/noop-persistence-provider.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export class NoopPersistenceProvider {
+	/**
+	 * Called when a document is bound to the provider.
+	 * In a real provider, this would load existing state for the doc from storage.
+	 *
+	 * @param docName - The name/identifier of the document
+	 * @param yDoc - The Y.js CRDT document instance
+	 */
+	bindState( docName: string, yDoc: CRDTDoc ) {
```
chriszarate:
```suggestion
	 * @param {string} _docName The name/identifier of the document
	 * @param {Y.Doc}  _ydoc    The Y.js CRDT document instance
	 */
	public bindState( _docName: string, _ydoc: Y.Doc ): void {
```

websocket-server/noop-persistence-provider.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { CRDTDoc } from '@wordpress/sync';
+
+/**
+ * A no-operation persistence provider for Y.js collaborative documents.
+ *
+ * This provider implements the persistence interface but doesn't actually
+ * persist any data to storage. It is mainly to ensure that, the websocket-server
+ * ejects documents from memory when there are no active connections.
+ */
+export class NoopPersistenceProvider {
```
chriszarate, line 13:
```suggestion
export class NoopPersistenceProvider {
	public provider = this;
```

websocket-server/noop-persistence-provider.ts
```diff
@@ -0,0 +1,37 @@
...
+
+	/**
+	 * Called when the document state should be persisted.
+	 * In a real provider, this would save the document state to storage.
+	 *
+	 * @param docName - The name/identifier of the document
+	 * @param yDoc - The Y.js CRDT document instance
+	 * @returns Promise that resolves when the write operation is complete
+	 */
+	writeState( docName: string, yDoc: CRDTDoc ) {
```
chriszarate:
```suggestion
	 * @param {string} _docName The name/identifier of the document
	 * @param {Y.Doc}  _ydoc    The Y.js CRDT document instance
	 * @returns Promise that resolves when the write operation is complete
	 */
	public writeState( _docName: string, _ydoc: Y.Doc ): Promise< void > {
```

websocket-server/package.json
```diff
@@ -20,6 +20,7 @@
     "@types/node": "22.18.6",
     "@types/ws": "8.5.0",
     "tsx": "4.20.5",
-    "typescript": "5.9.2"
+    "typescript": "5.9.2",
+    "@wordpress/sync": "1.31.0"
```
chriszarate: Remove and install `yjs`

websocket-server/tsconfig.json
```diff
@@ -11,10 +11,11 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "sourceMap": true,
+    "typeRoots": [ "../types", "./node_modules/@types" ],
```
chriszarate: Remove to keep package independent. This package is built separately in CI.

```suggestion
```

websocket-server/tsconfig.json
```diff
@@ -11,10 +11,11 @@
...
     "paths": {
       "@y/websocket-server/*": [ "./node_modules/@y/websocket-server/dist/src/*" ]
     }
   },
-  "include": [ "*.ts" ],
+  "include": [ "*.ts", "../types/*.d.ts" ],
```
chriszarate: Remove to keep package independent. This package is built separately in CI.

```suggestion
```

chriszarate (changes requested):
some cleanup needed

chriszarate:
> Implemented a delay after all the connections are closed in removing the doc from in-memory. We can see after this goes in if we want this

We might have to fork to implement this. 🤔 We could in theory "persist" the document in memory for the delay period and apply the changes if the user reconnects before the delay expires. But that would be messy and a more direct approach is probably better.

ingeniumed:
> > Implemented a delay after all the connections are closed in removing the doc from in-memory. We can see after this goes in if we want this
> 
> We might have to fork to implement this. 🤔 We could in theory "persist" the document in memory for the delay period and apply the changes if the user reconnects before the delay expires. But that would be messy and a more direct approach is probably better.

This was one of the reasons why I had originally forked the code, and added it within our websocket-server implementation. 

My recommendation would be to, to go ahead with the no-op persistence for now and see how it fares in production. 

Once we decide that we need a delay in removing/nuking the doc from in-memory, we can go at it in two ways:

- Attempt to push a fix upstream that allows for a function to be provided instead for eviction, when no persistence is set.
- Fork the code, if the above is not possible or takes too long.

I 100% wouldn't want to tweak the persistence provider in the way u mentioned because it could lead to a lot more foot gun situations, especially considering how our post_meta -> editor setup works. We don't want to mess that up, or cause problems there. I think we should rely on that until we are 100% certain that we want that delay.

websocket-server/noop-persistence-provider.ts
ingeniumed: Gah, I knew I was missing something obvious in getting this set with just a simple instantiation

websocket-server/noop-persistence-provider.ts
ingeniumed: Fair point, I've gone ahead and done that.

websocket-server/tsconfig.json
ingeniumed: Installing yjs renders this moot

websocket-server/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "sourceMap": true,
+    "typeRoots": [ "./node_modules/@types" ],
```
chriszarate: This is more restrictive than the default, which already includes this folder:

https://www.typescriptlang.org/tsconfig/#typeRoots


```suggestion
```

chriszarate (approved):

PR #48, opened 2025-09-29
Fix the plugin zip command

ingeniumed:
### Description

This'll ensure that we don't make another folder with the plugin's name in the zip file. So now the release workflow will actually work as expected


shekharnwagh (approved):
Thank you for fixing this ! 🚢

PR #55, opened 2025-10-03
Support relative positioning of cursors

alecgeatches:
See paired Gutenberg PR in https://github.com/Automattic/gutenberg/pull/41. If testing, please use both branches at the same time.

This PR utilizes [`Y.RelativePosition`](https://docs.yjs.dev/api/relative-positions) to draw user cursors instead of relying on an offset.

*Before*:

https://github.com/user-attachments/assets/a7c509d2-8926-4db6-928b-6ba7166bd48d

*After*:

https://github.com/user-attachments/assets/5f73a195-2673-49fb-b8e4-b7008a598b3c

## Not Done

Note that this PR does not do a few related things that will be covered in future PRs:

1. Now that we have relative positions we can correctly draw the other user's cursor, but we do not yet update that user's cursor. So if they type, it's still in the "before" position. This requires some intensive updating of selection that's still under experimentation, but it's a crucial part of this overall feature.
2. The `findBlockByClientId()` function is highly inefficient, recursing through the whole `Y.Doc` on each cursor update to find the relevant `Y.Text` object. It's not noticeably slow in testing, but for a large document with many cursors could be a bottleneck. This function should be updated to be a simple lookup. Here are some ideas for that:
    - Cache block client ID + attribute name to the relevant `Y.Text` object in `findBlockByClientId()`. We would need to be careful about invalidation when adding and removing blocks and `innerBlocks`.
    - Store clientID + attribute name at the root level of the `Y.Doc` to make lookup easy. Due to `innerBlocks`, recursion is currently necessary to find all `Y.Text` instances in the `Y.Doc`. If we're able to store a root-level map to `Y.Text` objects (or proxies to Y.Text objects stored in blocks) it would make the code a lot easier.
3. The UndoManager should be upgraded to [store relative cursor position](https://docs.yjs.dev/api/undo-manager#example-add-additional-information-to-the-stackitems) with changes.


src/utilities/user.ts
```diff
@@ -1,23 +1,5 @@
...
-		// Update this switch with any non-primitive fields added to UserState.
-		switch ( key ) {
-			case 'editorState':
-				return areEditorStatesEqual( userState1.editorState, userState2.editorState );
-
-			default:
-				return value === userState2[ key as keyof UserState ];
-		}
-	} );
+	return JSON.stringify( userState1 ) === JSON.stringify( userState2 );
```
alecgeatches: @chriszarate I changed this whole function to just use `JSON.stringify()`. I did this because it took me a while to track down a bug that caused excessive awareness updates. The cause was `areSelectionsEqual()` not being updated to correctly handle comparison of the new relative positions. Because `areUserStatesEqual()` was always returning `false`, this added a lot of redraws and noise when nothing was actually happening to the document. If the comparison code had been wrong in the other way (incorrectly returning `true`) it would have broken awareness updates for the new data type.

Overall, it does seem like an easy problem to encounter each time we add to user state or make a data type change. It's a minor footgun that can cause subtle performance issues and seems likely to happen again. Instead, this was changed to just use a `JSON.stringify()` comparison. It currently works for all data types in editor state, and unless we add something non-serializable to local state it should work well as a replacement. If we do need to come back to a property-by-property comparison, it should be easy to AI-up that code.

chriszarate: > it took me a while to track down a bug that caused excessive awareness updates. The cause was areSelectionsEqual() not being updated to correctly handle comparison of the new relative positions

Sorry about that! I should have coded it more defensively. More ideally it should `throw` an error in the `default` case instead of returning `false`:

https://github.com/Automattic/vip-real-time-collaboration/blob/e49403cc565928d79632c58c72be40e8c71c996a/src/utilities/selection.ts#L96

I think I would prefer this route instead, if possible. `JSON.stringify` has its own performance issues and potential footguns, such as unserializable properties that just get dropped. What do you think?

alecgeatches: > Sorry about that! I should have coded it more defensively.

This certainly isn't your fault. In this case, I changed the data type of `selection.cursorPosition`, so the non-`default` code still executed, it just started returning `false` where it previously hadn't because a a deeper data type changed. We just need to be really diligent about any changes to editor state, because false negatives will lead to incorrect UI and false positives will lead to subtle performance issues that get worse with more users. Or we should consider a different representation for editor states where the logic is simplified.

For now, I made a few changes:

1. In https://github.com/Automattic/vip-real-time-collaboration/pull/55/commits/7b095dc4de9c67d344ac1b802addfa11bfaa770b, changed comparison back to per-property as before. Note that relative positions are still compared via `JSON.stringify()` as this is [explicitly supported](https://docs.yjs.dev/api/relative-positions). Relative positions look like this: `{"type":{"client":305620264,"clock":1458},"tname":null,"item":{"client":305620264,"clock":1472},"assoc":0}`, so comparing them property-by-property would require careful iteration of the object internals, which feels like unnecessary coupling.
2. Removed a `throw` in https://github.com/Automattic/vip-real-time-collaboration/pull/55/commits/13d37ac89dc1e383fe869146ee96d15c2935f6e2 that I occasionally saw in testing and replaced it with gracefully not drawing a cursor instead.
3. Similar to your suggestion above, I changed the `default` case to log an error in https://github.com/Automattic/vip-real-time-collaboration/pull/55/commits/541a9f74e8d42d672cb7427ddbecc6543c0275ce. I've been trying to remove `throw`s in Awareness code unless it's something fatal. I don't imagine we'll have even more selection types come into play, but if we do and it causes a bug we'll at least have an error to look at.

src/utilities/selection.ts
```diff
@@ -194,3 +198,120 @@ export async function updateSelectionInEntityRecord(
...
+				areCursorPositionsEqual(
+					selection1.cursorEndPosition,
+					( selection2 as SelectionInMultipleBlocks ).cursorEndPosition
+				)
+			);
+		case SelectionType.WholeBlock:
+			return selection1.blockId === ( selection2 as SelectionWholeBlock ).blockId;
+
+		default:
+			console.error( 'Unable to compare selection types:', selection1, selection2 );
```
chriszarate: Use our logger here

src/utilities/selection.ts
```diff
@@ -194,3 +198,120 @@ export async function updateSelectionInEntityRecord(
...
+			if ( innerBlock ) {
+				return innerBlock;
+			}
+		}
+	}
+
+	return null;
+}
+
+export function areSelectionsEqual(
```
chriszarate, line 252: it would be great to have unit tests for these equality functions. AI should excel at writing them.

src/awareness-manager.ts
```diff
@@ -259,44 +269,65 @@ export class AwarenessManager {
...
-						this.setLocalStateField( 'editorState', pendingEditorState );
-						pendingEditorState = null;
-					}
-					awarenessCursorTimeout = null;
-				}, AWARENESS_CURSOR_UPDATE_DEBOUNCE_IN_MS );
-			}
 		} );
 	}
 
+	private updateSelectionState(
```
chriszarate, line 297: we might think about pulling out the selection logic into a separate class or concern. getting heavy

src/utilities/selection.ts
alecgeatches: Addressed in https://github.com/Automattic/vip-real-time-collaboration/pull/55/commits/78889f54f973e517f835e4c3e7989389b22acd8c!

src/utilities/selection.ts
alecgeatches: Good idea, added tests in https://github.com/Automattic/vip-real-time-collaboration/pull/55/commits/f53cc8beecc5cf822c4dedc453c95d5c64fd5353.

src/awareness-manager.ts
alecgeatches: This file is definitely getting large, will attempt to break it up in future changes.

PR #58, opened 2025-10-07
Fix awareness UI for non-iframe block editor

maxschmeling:
Fixes VIP-1940

See https://github.com/Automattic/gutenberg/pull/42 for the required Gutenberg changes

This uses the new `<BlockCanvasCover>` SlotFill in the Gutenberg branch to more reliably render the awareness UI over both the iframe based editor and the non-iframe based editor.

To test, activate a non-block based theme and add a block with `apiVersion: 2` then switch between the desktop view (non-iframe editor) and tablet view (iframe editor).


vip-real-time-collaboration.php
```diff
@@ -40,7 +40,7 @@
 
 // Examples (must be manually built):
 // require_once __DIR__ . '/examples/local-updates-block/local-updates-block.php';
-// require_once __DIR__ . '/examples/post-meta-block/post-meta-block.php';
+require_once __DIR__ . '/examples/post-meta-block/post-meta-block.php';
```
ingeniumed: Should this be left enabled? I assume it was just for testing.

package.json
```diff
@@ -85,6 +86,7 @@
     "yjs": "13.6.27"
   },
   "dependencies": {
+    "@wordpress/private-apis": "^1.31.0",
```
ingeniumed: Where is this used?

ingeniumed:
Got some type failures related to the new SlotFill

examples/post-meta-block/src/subtitle/block.json
```diff
@@ -1,6 +1,6 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/block.json",
-	"apiVersion": 3,
+	"apiVersion": 2,
```
chriszarate: why?

maxschmeling: This is for testing. I'm pulling these changes out now.

package.json
maxschmeling: Good catch. From earlier attempts. Removed in c65b789

vip-real-time-collaboration.php
maxschmeling: This enabled the testing (if you build that example) but it needs removed for the PR. Done here e833f61

examples/post-meta-block/src/subtitle/block.json
maxschmeling: e833f61

src/components/rtc-settings-panel.tsx
```diff
@@ -67,69 +72,82 @@ export function RTCSettingsPanel() {
...
-					label="Enable cursors"
-					checked={ isCursorsEnabled }
-					onChange={ ( enabled: boolean ) => {
-						handleToggleCursors( enabled );
-					} }
-				/>
+		<>
+			{ isAvatarsEnabled && (
+				<BlockCanvasCover.Fill>
+					<Avatars />
```
ingeniumed, line 74: The debug tools need to be added back in here. They got removed from it's previous location, and wasn't added back. That's why it's not working

ingeniumed: Debug tools are working again, can confirm

ingeniumed: Also the disconnection UI is working again as well

shekharnwagh (approved):

PR #60, opened 2025-10-08
Fix fatal error when experiments option is set to empty string

chriszarate:
Fix fatal error (due to type hint) when experiments option is set to empty string. Add integration tests to cover possible option values.

### DIFF
--- a/inc/Compatibility/Compatibility.php
+++ b/inc/Compatibility/Compatibility.php
@@ -47,9 +47,7 @@ final class Compatibility {
 			$experiments = [];
 		}
 
-		if ( array_key_exists( 'gutenberg-sync-collaboration', $experiments ) ) {
-			unset( $experiments['gutenberg-sync-collaboration'] );
-		}
+		unset( $experiments['gutenberg-sync-collaboration'] );
 
 		// Do not enable on Site Editor.
 		if ( 'site-editor.php' === $pagenow ) {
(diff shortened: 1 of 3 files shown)


inc/Compatibility/Compatibility.php
```diff
@@ -40,12 +40,20 @@ public static function admin_notices(): void {
...
 	 * @psalm-suppress PossiblyUnusedReturnValue Psalm does not detect usage via add_filter.
 	 */
-	public function enable_sync_collaboration_experiment( array|false $experiments ): array|false {
+	public function enable_sync_collaboration_experiment( mixed $experiments ): array|false {
 		global $pagenow;
 
 		// Do not enable on Site Editor.
 		if ( 'site-editor.php' === $pagenow ) {
-			return $experiments;
+			if ( is_array( $experiments ) ) {
```
ingeniumed: What do you think about simplifying this as follows:

- re-initializing experiments as an empty array when it's not an array, at the very start of the function
- The if checks becomes, unsetting the key if the site-editor is the page being loaded, otherwise the key is set to true.

So it would look as follows:

```php
global $pagenow;

if ( ! is_array( $experiments ) ) {
 // set $experiments to be an array
}

if ( 'site-editor.php' === $pagenow ) {
unset( $experiments['gutenberg-sync-collaboration'] );
} else {
$experiments['gutenberg-sync-collaboration'] = true;
}

return $experiments
```

The unset command doesn't need to worry about the key existing or not, as it won't error out if its not present.

ingeniumed (approved):
Just the one change that's related to simplifying the code down. The actual logic is good, its more to see how we could improve the code further.

inc/Compatibility/Compatibility.php
chriszarate: Updated

ingeniumed (approved):

PR #61, opened 2025-10-09
Implement removal of sync settings field in Gutenberg

pkevan:
Remove the sync collaboration settings field from the Gutenberg experiments page to avoid users thinking it can be disabled.

Follow-up to: #60


pkevan:
Screenshot 2025-10-09 at 17 16 57

Could also have something like this included.

chriszarate (approved):

PR #63, opened 2025-10-13
Add a PHP and WP version check

ingeniumed:
### Description

We don't currently have any checks to enforce the WP/PHP requirements that we have in place, which is of 8.2 and 6.7. This change will implement that feature, such that an admin notice will be visible informing a user about an error.

There are two key differences compared to how the Governance plugin did this:

- Use `wp_admin_notice` as our minimum version is 6.7 not 6.4. We could go for the older way, but I think it's fine imo.
- WP's version.php is required to ensure that if plugins override the wp_version then it'll still work.

wp_get_wp_version cannot be used btw, as that's 6.7+ which ironically would not work for this problem.

### Testing

- Either load an earlier version of WP/PHP and load the plugin
- Or, flip the signs and test it on a newer WP/PHP

CleanShot 2025-10-13 at 15 15 08

### DIFF
--- a/vip-real-time-collaboration.php
+++ b/vip-real-time-collaboration.php
@@ -28,6 +28,37 @@ if ( defined( 'VIP_REAL_TIME_COLLABORATION__LOADED' ) ) {
 }
 
 define( 'VIP_REAL_TIME_COLLABORATION__LOADED', true );
+
+/** @psalm-suppress PossiblyFalseArgument */
+if ( version_compare( phpversion(), '8.2', '<' ) ) {
+	add_action( 'admin_notices', function (): void {
+		wp_admin_notice(
+			__(
+				'The VIP Real-Time Collaboration plugin requires PHP 8.2+. The VIP Real-Time Collaboration plugin has been disabled.',
+				'vip_real_time_collaboration'
+			),
+			[ 'type' => 'error' ]
+		);
+	}, 10, 0 );
+	return;
+}
+
+/** @psalm-suppress InvalidGlobal */
+global $wp_version;
+/** @psalm-suppress MixedArgument */
+if ( version_compare( $wp_version, '6.7', '<' ) ) {
+	add_action( 'admin_notices', function (): void {
+		wp_admin_notice(
+			__(
+				'The VIP Real-Time Collaboration plugin requires WordPress 6.7+. The VIP Real-Time Collaboration plugin has been disabled.',
+				'vip_real_time_collaboration'
+			),
+			[ 'type' => 'error' ]
+		);
+	}, 10, 0 );
+	return;
+}
+
 define( 'VIP_REAL_TIME_COLLABORATION__PLUGIN_ROOT', __FILE__ );
 define( 'VIP_REAL_TIME_COLLABORATION__PLUGIN_DIRECTORY', untrailingslashit( plugin_dir_path( __FILE__ ) ) );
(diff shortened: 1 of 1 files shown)


vip-real-time-collaboration.php
```diff
@@ -60,3 +64,48 @@
...
+		}, 10, 0 );
+		return false;
+	}
+
+	global $wp_version;
+
+	// Account for plugins overriding the $wp_version global, look at Gutenberg.php for reference.
+	/** @psalm-suppress MissingFile */
+	// This is a built-in WordPress file, so we can ignore the warning here.
+	require ABSPATH . WPINC . '/version.php';
```
maxschmeling: This matches how `gutenberg.php` does it. A little safer. Unless there's a reason we should `require`.
```suggestion
	include ABSPATH . WPINC . '/version.php';
```

vip-real-time-collaboration.php
```diff
@@ -60,3 +64,48 @@
...
+				),
+				[ 'type' => 'error' ]
+			);
+		}, 10, 0 );
+		return false;
+	}
+
+	global $wp_version;
+
+	// Account for plugins overriding the $wp_version global, look at Gutenberg.php for reference.
```
maxschmeling: Very minor, but I'm guessing it capitalized on its own since it did it to me too when I typed it.
```suggestion
	// Account for plugins overriding the $wp_version global, look at gutenberg.php for reference.
```

maxschmeling (approved):

vip-real-time-collaboration.php
```diff
@@ -60,3 +64,48 @@
...
+/**
+ * Verify that we can initialize the VIP Real-Time Collaboration plugin.
+ *
+ * @global string $wp_version The WordPress version string.
+ *
+ * @return bool true if the plugin can load, false otherwise.
+ */
+function vip_real_time_collaboration_pre_init(): bool {
+	$php_version = phpversion();
+	if ( is_string( $php_version ) && version_compare( $php_version, '8.2', '<' ) ) {
```
chriszarate, line 77: Can you use `validate_plugin_requirements`?

https://developer.wordpress.org/reference/functions/validate_plugin_requirements/

chriszarate (changes requested):

vip-real-time-collaboration.php
chriszarate: I guess we can't since it won't be in `WP_PLUGIN_DIR`.

chriszarate (approved):

vip-real-time-collaboration.php
ingeniumed: Yip, that's why I didn't consider it. We ran into that when developing the governance plugin

vip-real-time-collaboration.php
ingeniumed: Require will halt the execution if its unable to get the version.php loaded, while include will continue on. Noticed in core that, they used require - https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop%20ABSPATH%20.%20WPINC%20.%20%27%2Fversion.php%27%3B&type=code so I ended up switching it to that as well.

Happy to match it with Gutenberg, can re-visit this if we need to.

vip-real-time-collaboration.php
ingeniumed: Good catch, it was auto-corrected there.

PR #65, opened 2025-10-14
Move <Avatars> to the header

maxschmeling:
* Requires https://github.com/Automattic/gutenberg/pull/46

More closely aligns with the latest design work by moving the avatars out of the editor area and into the header.

Refs VIP-2077

### DIFF
--- a/src/components/avatars.tsx
+++ b/src/components/avatars.tsx
@@ -24,22 +24,17 @@ export function Avatars() {
 	const remainingUsers = activeUsers.slice( 3 );
 	const remainingUsersText = remainingUsers.map( userState => userState.name ).join( ', ' );
 
-	return (
-		<div className="vip-real-time-collaboration-avatars-container">
-			<div className="vip-real-time-collaboration-avatars">
-				{ visibleUsers.map( userState => (
-					<Avatar key={ userState.clientId } userState={ userState } showUserColorBorder={ true } />
-				) ) }
+	return visibleUsers.length > 1 ? (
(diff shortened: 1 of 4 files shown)


src/components/avatars.tsx
```diff
@@ -24,22 +24,17 @@ export function Avatars() {
...
-	return (
-		<div className="vip-real-time-collaboration-avatars-container">
-			<div className="vip-real-time-collaboration-avatars">
-				{ visibleUsers.map( userState => (
-					<Avatar key={ userState.clientId } userState={ userState } showUserColorBorder={ true } />
-				) ) }
+	return visibleUsers.length > 1 ? (
+		<>
+			{ visibleUsers.map( userState => (
+				<Avatar key={ userState.clientId } userState={ userState } showUserColorBorder={ false } />
```
alecgeatches, line 30: Any reason you changed `showUserColorBorder={ true }` to `false`? I think the border colors are helpful to have here.

alecgeatches (approved):
I love the new look with rounded squares! Here's a screenshot for anybody else reviewing:

Screenshot 2025-10-14 at 1 31 50 PM

src/components/avatars.tsx
maxschmeling: Just to match the designs.

I think it makes sense to have them there, but they're smaller than they were before and I'm not sure it's easy to match them up.

CleanShot 2025-10-14 at 15 08 33@2x

PR #68, opened 2025-10-20
Check if the gutenberg plugin matches the version we expect

ingeniumed:
### Description

We don't currently check if the Gutenberg plugin being used matches the version expected for RTC. This change verifies if the version matches what's expected which is set to 21.9.0 (RC for WordPress 6.9), and only does so if its not development mode.

It's not 100% accurate to say that 21.9.0 is what we need because there's more pieces in our special branch but we have to start somewhere. I'm hoping we can bumping this number up with each release that contains more pieces from that special branch.

### Testing

Loading the normal version of Gutenberg will trigger this error because 21.9.0 hasn't been released yet.

### Screenshot

CleanShot 2025-10-20 at 14 50 22


inc/Compatibility/Compatibility.php
```diff
@@ -70,13 +68,30 @@ public static function get_supported_post_types(): array {
...
+		// For production builds, this is removed entirely.
+		if ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) && constant( 'GUTENBERG_DEVELOPMENT_MODE' ) ) {
+			return true;
+		}
+
+		// If its not development mode, and the version is set, check the version.
+		if ( defined( 'GUTENBERG_VERSION' ) && is_string( constant( 'GUTENBERG_VERSION' ) ) ) {
+			/** @var string $gutenberg_version */
+			$gutenberg_version = constant( 'GUTENBERG_VERSION' );
+			return version_compare( $gutenberg_version, '21.9.0', '>=' );
```
ingeniumed: Given 21.9.0 will be where our changes are, I've set this to 21.9.0. We can have this be the floor for our plugin, so that we can at least have some guarantee that the plugin will work.

ingeniumed: Note: 21.9.0 will be released this week.

inc/Compatibility/Compatibility.php
```diff
@@ -70,13 +68,30 @@ public static function get_supported_post_types(): array {
...
-		return defined( 'IS_GUTENBERG_PLUGIN' ) && constant( 'IS_GUTENBERG_PLUGIN' );
+		// Gutenberg plugin isn't active.
+		if ( ! defined( 'IS_GUTENBERG_PLUGIN' ) || ! constant( 'IS_GUTENBERG_PLUGIN' ) ) {
+			return false;
+		}
+
+		// Need to ensure that development mode bypasses this check.
+		// For dev builds, this is defined in the Gutenberg plugin's main file.
+		// For production builds, this is removed entirely.
+		if ( defined( 'GUTENBERG_DEVELOPMENT_MODE' ) && constant( 'GUTENBERG_DEVELOPMENT_MODE' ) ) {
```
ingeniumed: This'll ensure that we don't check the version when Gutenberg is in development mode. I did it by checking Gutenberg in production vs development mode, and looking at the Gutenberg.php file. I initially checked SCRIPT_DEBUG, but that wasn't 100% accurate, if you look at the auto generated parts of the Gutenberg.php.

Happy to get feedback on this.

inc/Compatibility/Compatibility.php
ingeniumed, line 16: I originally had this GB check in the new pre_init method, but then I wondered if we might have some trouble if the GB plugin hadn't loaded yet? This method is called via `Compatibility::should_plugin_load()` in `add_action( 'plugins_loaded'` which reduce the likelihood of such an error. Hence why it's here.

pkevan:
At the moment, I guess it isn't fully accurate that the plugin required that version of Gutenberg - should we caution against calling out a specific version or at least mention it's a special modified version for this plugin, until we're at that point?

ingeniumed:
> At the moment, I guess it isn't fully accurate that the plugin required that version of Gutenberg - should we caution against calling out a specific version or at least mention it's a special modified version for this plugin, until we're at that point?

My main concern with originally putting up such a PR was targeting a fix with the fork in mind, rather than the core GB plugin. It's fair that, even with 21.9.0 we would still need our special branch. But, we have to start somewhere imo.

That said, I'm hoping with each subsequent release after that we can keep bumping this up as there'd be less and less in that special branch.

Hopefully that answers what you meant.

pkevan:
> > At the moment, I guess it isn't fully accurate that the plugin required that version of Gutenberg - should we caution against calling out a specific version or at least mention it's a special modified version for this plugin, until we're at that point?
> 
> My main concern with originally putting up such a PR was targeting a fix with the fork in mind, rather than the core GB plugin. It's fair that, even with 21.9.0 we would still need our special branch. But, we have to start somewhere imo.
> 
> That said, I'm hoping with each subsequent release after that we can keep bumping this up as there'd be less and less in that special branch.
> 
> Hopefully that answers what you meant.

Yeah, I understand the intent, my concern was mainly around giving a not fully true message - perhaps we can link to some documentation instead that makes it clearer, and which could follow whichever is true at that point in time as we progress to having it run via official GB versions?

ingeniumed:
> Yeah, I understand the intent, my concern was mainly around giving a not fully true message - perhaps we can link to some documentation instead that makes it clearer, and which could follow whichever is true at that point in time as we progress to having it run via official GB versions?

I think what I have added works for now, since we are still in the beta phase. As we get closer to a full release, there'll be wpvip documentation and we'd have 100% of our changes within the GB plugin so we'd be able to truly say that x version is needed.

ingeniumed:
Noting that with the release of 21.9.0 this is good to be merged in.

inc/Compatibility/Compatibility.php
```diff
@@ -70,13 +70,30 @@ public static function get_supported_post_types(): array {
...
+			return true;
+		}
+
+		// If its not development mode, and the version is set, check the version.
+		if ( defined( 'GUTENBERG_VERSION' ) && is_string( constant( 'GUTENBERG_VERSION' ) ) ) {
+			/** @var string $gutenberg_version */
+			$gutenberg_version = constant( 'GUTENBERG_VERSION' );
+			return version_compare( $gutenberg_version, '21.7.0', '>=' );
```
chriszarate: Why does this compare against 21.7.0 while the copy says 21.9.0+? Should they both pull from a constant?

chriszarate:
Is there a way to do feature detection instead of version comparison? My main worry is that this gets forgotten about.

ingeniumed:
> Is there a way to do feature detection instead of version comparison? My main worry is that this gets forgotten about.

I couldn't come up with a reliable and trusted way to do feature detection on the PHP side that'd give us the confidence to include it here. Since we are starting to make PRs upstream, I'd lean towards sticking with the version instead. Eventually, we should be able to say with confidence that if you have x version then you have everything in place to get RTC working.

inc/Compatibility/Compatibility.php
ingeniumed: I'm going to provide a constant at the top, rather than one like the websocket url that can be overriden. I think the constant in the code should be fixed, as it means that GB version is tied to that code. It's not exactly accurate since there's some changes in our wpvip/rtc-plugin but still.

ingeniumed:
Maybe [GUTENBERG_GIT_COMMIT](https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/vip-integrations/gutenberg/gutenberg.php#L17C10-L17C30) until such time that we use only the trunk branch, without any special changes?

That'd be specific to us. My only concern would be ensuring that we coordinate this with our RTC plugin.

The current approach is a bit less riskier and easier:

- Update GB's version
- Let that release go through
- Update RTC's minimum expected GB version immediately after
- Next week's release would include the RTC change

This is just for the mu-plugins/integrations side tbh.

inc/Compatibility/Compatibility.php
ingeniumed: This should be 21.9.0. I was waiting for it to roll out before I updated the version check as well. 

It's been updated, and I've added a class constant to track this.

pkevan:
What if we used the integration center API? `wpvip_get_integration_info` might solve this for us?

See: https://github.com/Automattic/vip-go-mu-plugins/pull/6527/

ingeniumed:
I'm reserving that type of a check for the code that'd go into mu-plugins. I'm trying to keep this check VIP specific free because this plugin could run on non-VIP hosts.

I should have mentioned that, but that's a goal I'm keeping mind. This plugin isn't meant to be VIP specific so our code shouldn't be VIP specific either. The mu-plugins code can be VIP specific, so that can take advantage of the API you linked when trying to load the integration.

maxschmeling:
What if we just embed a constant or something into our fork that we can check so that we know we're always running our fork? If it's deployed through the integration center we'll know it's the latest, so we really just want to check that it's *our* Gutenberg and not one that has been installed from the Gutenberg release or something.

ingeniumed:
> What if we just embed a constant or something into our fork that we can check so that we know we're always running our fork? If it's deployed through the integration center we'll know it's the latest, so we really just want to check that it's _our_ Gutenberg and not one that has been installed from the Gutenberg release or something.

That could work though we'd need to make sure to flag that we don't want to let that go in permanently. 

I am a bit hesitant because I'd like to get all our PRs into core and then drop the fork as quickly as we can imo. We would need to make sure that whenever we do switch away from that fork, we drop that check and only release Gutenberg post that release.

inc/Compatibility/Compatibility.php
```diff
@@ -27,7 +33,7 @@ public static function admin_notices(): void {
 		if ( ! self::is_websocket_url_defined() ) {
 			wp_admin_notice(
 				__(
-					'The WebSocket URL has not been configured. The VIP Real-Time Collaboration plugin has been disabled.',
+					'The VIP Real-Time Collaboration plugin requires a WebSocket URL to be configured. The VIP Real-Time Collaboration plugin has been disabled.',
```
chriszarate: Why this (repetitive) edit?

inc/Compatibility/Compatibility.php
```diff
@@ -70,13 +76,30 @@ public static function get_supported_post_types(): array {
 	}
 
 	/**
-	 * Check if the Gutenberg plugin is active.
-	 *
-	 * TODO: Check GUTENBERG_VERSION in production to ensure it is running a
-	 * compatible version.
+	 * Check if the Gutenberg plugin is active, and meets the minimum version requirement.
```
chriszarate:
```suggestion
	 * Check if the Gutenberg plugin is active and meets the minimum version requirement.
```

inc/Compatibility/Compatibility.php
```diff
@@ -70,13 +76,30 @@ public static function get_supported_post_types(): array {
...
 	private static function is_gutenberg_plugin_active(): bool {
-		return defined( 'IS_GUTENBERG_PLUGIN' ) && constant( 'IS_GUTENBERG_PLUGIN' );
+		// Gutenberg plugin isn't active.
+		if ( ! defined( 'IS_GUTENBERG_PLUGIN' ) || ! constant( 'IS_GUTENBERG_PLUGIN' ) ) {
+			return false;
+		}
+
+		// Need to ensure that development mode bypasses this check.
+		// For dev builds, this is defined in the Gutenberg plugin's main file.
+		// For production builds, this is removed entirely.
```
chriszarate:
```suggestion
		// This constant is only defined for Gutenberg development builds.
```

chriszarate:
> What if we just embed a constant or something into our fork that we can check so that we know we're always running our fork?

I think this is a good approach and less likely than the current approach to fail via neglect. We can mark the PR in our branch as `[no-pr][not-for-trunk]`.

ingeniumed:
PR: https://github.com/WordPress/gutenberg/pull/73684

This changes our custom GB fork to add in a constant - IS_CUSTOM_GUTENBERG_FORK in line with our discussion above.

CleanShot 2025-12-02 at 14 23 20

inc/Compatibility/Compatibility.php
```diff
@@ -17,7 +17,7 @@ public static function admin_notices(): void {
 		if ( ! self::is_gutenberg_plugin_active() ) {
 			wp_admin_notice(
 				__(
-					'The Gutenberg plugin has not been installed. The VIP Real-Time Collaboration plugin has been disabled.',
+					'The correct Gutenberg plugin has not been installed. The VIP Real-Time Collaboration plugin has been disabled.',
```
ingeniumed: I needed to communicate that the problem is that, our custom GB plugin has not been installed. Hence why I used correct here, and also avoided saying fork or VIP.

inc/Compatibility/Compatibility.php
ingeniumed: Fixed

ingeniumed:
Note: The integration tests will pass once the Gutenberg PR is merged into our branch.

inc/Compatibility/Compatibility.php
pkevan:
```suggestion
					'A compatible Gutenberg plugin is not installed so the VIP Real-Time Collaboration plugin has been disabled.',
```

pkevan: maybe like the above?

ingeniumed: I like that a lot!

ingeniumed:
This PR will be closed off as this approach is not the path we want to go down at the moment.

Instead, we are going to flag the version and the commit sha in the plugin settings so it's easily visible.

PR #70, opened 2025-10-21
Disable the code editor

ingeniumed:
### Description

The code editor is one possible way to trigger block recovery errors within the editor. It's also a way to turn valid block markup, into an invalid one. The code editor will now be disabled, in the event that the post supports collaboration. In all other cases, the code editor will be allowed.

CleanShot 2025-10-21 at 13 53 24

This'll supersede the closed PR in Gutenberg - https://github.com/Automattic/gutenberg/pull/47

### Testing

- Open up a post
- Try to get into the code editor mode
- That shouldn't be allowed with this change, while in trunk it would be.


chriszarate:
Why is this in `CrdtPersistence`?

ingeniumed:
> Why is this in `CrdtPersistence`?

I wanted it in the editor folder tbh, that's really why. I didn't move it to its own file as it'd be another file with just one action/filter in there.

Lemme push up another file.

chriszarate:
What about `Overrides`?

ingeniumed:
I originally had it there, but that didn't feel right because we are changing a setting within the editor that'd change its UI. We aren't overriding something in the editor.

I'm thinking EditorUI under editor makes more sense? That way, anything related to the UI of the editor could go here as well.

ingeniumed:
Pushed that up, lemme know what you think.

I was torn between editorUI, editorSettings, editorTweaks, and a million other names that implied we had changed the UI of the editor via settings.

Since its settings right now, and could expand to some other method down the line I figured editorUI made the most sense.

pkevan:
Just wondering if we should allow the code editor for certain roles, or with a capability? What happens when a block become corrupt through a no-fault edit, how would this recover?

chriszarate:
> What happens when a block become corrupt through a no-fault edit, how would this recover?

Perhaps we need a way to disable it only for collaborative sessions and not in "single-user mode."

Question: Could we allow the code editor in collaborative sessions, but in "read-only" mode?

ingeniumed:
> Could we allow the code editor in collaborative sessions, but in "read-only" mode?

This specific way is an on or off, no middle ground. I don't think there's a read-only mode, but let me re-confirm this.

> disable it only for collaborative sessions and not in "single-user mode."

One challenge here is that, this filter is a PHP one and AFAIK we can only detect the "single-user mode" via JS since we need access to Yjs.

Lemme experiment with this, to see if I can find something using what we did in the governance plugin.

ingeniumed:
I got the code-editor to be read only! I have to clean the code up before I can push it up.

ingeniumed:
Closing this in favour of https://github.com/Automattic/vip-real-time-collaboration/pull/73. The package.json and type changes seem to have messed up the branch.

PR #72, opened 2025-10-22
Fix type errors

ingeniumed:
### Description

There's type failures happening in #73, that I've attempted to fix. I notice there were some methods that were unused that were starting to cause type failures so I decided to address that.

There's test errors happening locally that don't seem to be replicated here.

PR #73, opened 2025-10-22
Add a read only code editor mode

ingeniumed:
### Description

#70 had to be closed due to some package conflicts. 

Now, we are able to add a read-only mode for the code editor with the option of a filter that can be used to tweak when this occurs.

CleanShot 2025-10-22 at 11 50 31

### Testing

- Load a post
- Go to the code editor, instead of the visual editor
- Watch how its disabled so you can see it but not edit it


ingeniumed:
https://github.com/Automattic/vip-real-time-collaboration/pull/70#issuecomment-3429811117

Noting this comment here as that's the one I addressed.

ingeniumed:
The package-lock.json changes are weird. I think they are related to #71, and I haven't been able to reduce em no matter what I do. So I've left it with the intention that we will be bumping up the WP packages soon .

ingeniumed:
Opened https://github.com/Automattic/vip-real-time-collaboration/pull/72 to fix the type and test problems.

pkevan:
While this does stop editing in the code editor, it's trivial to override the behaviour by removing the `readonly` element from the textarea. If you're willing to be editing the code, I suspect this is not really going to deter you.


https://github.com/user-attachments/assets/4cfadfb1-6f21-4346-b004-5a6aabbd03b2

chriszarate:
> While this does stop editing in the code editor, it's trivial to override the behaviour

This doesn't bother me too much. There's plenty trouble you can get into if you override DOM-based protections like `readonly` or remove interaction-blocking overlays (like the post lock modal).

I do wonder if we should make this change in Gutenberg to provide a formal client-side avenue for disabling the code editor or making it read-only.

ingeniumed:
> I do wonder if we should make this change in Gutenberg to provide a formal client-side avenue for disabling the code editor or making it read-only.

That's what I'd like as the end state, but that'll take a while to propose and get into GB.

I think we should get it into our RTC plugin in the meantime - this PR basically. 

We can propose something for Gutenberg, and once that's in we can replace our solution with that.

maxschmeling:
I pushed fae2938277801218f36d8af171f306f949f0dff5 which just adds a more explicit check on the element. You can revert it if you think it's worse. It's probably not necessary, but it's more explicit about typing.

maxschmeling (approved):
I'm ok merging this for now, but we should try to figure out the long term solution in Gutenberg sooner rather than later.

ingeniumed:
Thanks for adding in that fix. I've also corrected the linting errors, and fixed the package-lock.json.

The check-types are a problem. I've started an internal convo on this, can chime in there @maxschmeling.

ingeniumed:
Okay, I've finally fixed the problem.

I had to mark data, e2e-test-utils-playwright, and editor as exact versions to be pulled in or else with the addition of `edit-post` and the new updates that went out to those packages the latest ones were being pulled in. 

The latest packages seem to have a type problem which is causing our `check-types` to start failing from what I can tell. Gutenberg did shift to a new build pipeline so I wonder if we need to tweak how we look for types?

PR #76, opened 2025-10-23
Hide lock modal based on user access to feature

chriszarate:
Hide lock modal based on user access to collaborative editing feature. Companion to https://github.com/WordPress/gutenberg/pull/72644.

If we want to allow users with access to the feature to exist alongside users without access, we need a way to selectively enforce the post lock.

If both the current user and the lock owner have access to the collaborative editing feature, then we can safely ignore the post lock (but continue updating it in the background). Otherwise, defer to the current behavior of the post lock.

## Testing instructions

Use this example cap filter to test this PR by removing the `edit_with_others` cap from a user that would otherwise have it based on their role.

```php
// Filter to exclude user ID 2 from the edit_with_others capability
add_filter( 'user_has_cap', function ( array $allcaps, array $caps, array $args, WP_User $user ): array {
	// Check if we're checking for the 'edit_with_others' capability
	if ( in_array( 'edit_with_others', $caps, true ) && 2 === $user->ID ) {
		$allcaps['edit_with_others'] = false;
	}

	return $allcaps;
}, 10, 4 );
```

Then:

1. Check out this branch.
2. Check out the Gutenberg branch from [this PR](https://github.com/WordPress/gutenberg/pull/72644).
1. Open a post with user 1.
3. Open the same post with user 2.
4. Take over the post.
5. Confirm that user 1 sees a post lock dialog (after the heartbeat completes).
6. Confirm the reverse.
7. Confirm that two users with the cap intact can collaborate without a post lock dialog.

### DIFF
--- a/inc/Auth/SyncPermissions.php
+++ b/inc/Auth/SyncPermissions.php
@@ -10,6 +10,8 @@ defined( 'ABSPATH' ) || exit;
  * Handles permission checking for sync objects using custom sync capabilities
  */
 final class SyncPermissions {
+	const CAP_NAME = 'edit_with_others';
+
 	/**
 	 * Initialize custom sync capabilities.
 	 * Sets up meta capability mapping and role capabilities.
@@ -110,7 +112,7 @@ final class SyncPermissions {
 		// Check sync_post capability (will be mapped to edit_post via map_meta_cap)
 		/** @var bool|WP_Error $can_sync_post */
 		$can_sync_post = true;
-		if ( ! current_user_can( 'sync_post', $parsed_post_id ) ) {
+		if ( ! current_user_can( self::CAP_NAME, $parsed_post_id ) ) {
 			$can_sync_post = new WP_Error(
 				'insufficient_sync_permissions',
 				__( 'You do not have permission to sync this content', 'vip-real-time-collaboration' )
@@ -194,8 +196,8 @@ final class SyncPermissions {
 
 		foreach ( $roles_to_update as $role_name ) {
 			$role = get_role( $role_name );
-			if ( $role && ! $role->has_cap( 'sync_post' ) ) {
-				$role->add_cap( 'sync_post' );
+			if ( $role && ! $role->has_cap( self::CAP_NAME ) ) {
+				$role->add_cap( self::CAP_NAME );
 			}
 		}
 	}
--- a/inc/Overrides/Overrides.php
+++ b/inc/Overrides/Overrides.php
@@ -21,9 +21,6 @@ final class Overrides {
 	 * Constructor to initialize the overrides.
 	 */
 	public function __construct() {
-		// Allow multiple users to see the edit post screen.
-		add_filter( 'show_post_locked_dialog', '__return_false' );
-
 		// Ensure that the _edit_lock meta key is never returned, effectively disabling the post lock functionality just for revisions.php.
 		add_filter( 'get_post_metadata', [ $this, 'filter_post_meta' ], 10, 3 );
 	}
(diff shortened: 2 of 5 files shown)


inc/Overrides/Overrides.php
```diff
@@ -21,9 +21,6 @@ final class Overrides {
 	 * Constructor to initialize the overrides.
 	 */
 	public function __construct() {
-		// Allow multiple users to see the edit post screen.
-		add_filter( 'show_post_locked_dialog', '__return_false' );
```
pkevan, line 25: this was left in, as this only occurs when not using the block editor.

inc/Auth/SyncPermissions.php
```diff
@@ -110,7 +112,7 @@ private static function check_post_sync_permissions(
 		// Check sync_post capability (will be mapped to edit_post via map_meta_cap)
 		/** @var bool|WP_Error $can_sync_post */
 		$can_sync_post = true;
-		if ( ! current_user_can( 'sync_post', $parsed_post_id ) ) {
+		if ( ! current_user_can( self::CAP_NAME, $parsed_post_id ) ) {
 			$can_sync_post = new WP_Error(
 				'insufficient_sync_permissions',
 				__( 'You do not have permission to sync this content', 'vip-real-time-collaboration' )
```
pkevan, line 118: should this be better described now the cap has changed from `sync_post`?

"You do not have permission to edit with other users"?

inc/Overrides/Overrides.php
chriszarate: Meaning the Site Editor (Appearance > Editor)? Or are there other cases?

inc/Auth/SyncPermissions.php
ingeniumed: Good opportunity for

"I'm afraid I can't let you do that ____"

ingeniumed:
I keep getting an error with this, and the corresponding Gutenberg branch that breaks RTC entirely. There's no way for me to get rid of the error, or to get RTC working again.

I tried this with users that have the `edit_with_others` capability, and those without. Both situations result in the exact same error. I've attached the react minified error.

CleanShot 2025-10-27 at 14 47 41

CleanShot 2025-10-27 at 14 56 11

I've run a `distclean` on the Gutenberg branch, and cleaned out the RTC plugin's build directory as well. I've also tried to wipe away the vip dev-env and the same error comes up.

What I'm able to do though is collaborate on a post the normal way, which is to take over a post and see the lock show up on the other user.

inc/Overrides/Overrides.php
pkevan: Possibly, but I always struggled to trigger it happening. 

TBH - given the intent for all this work is within the block editor maybe it's OK to just remove it.

pkevan:
> I keep getting an error with this, and the corresponding Gutenberg branch that breaks RTC entirely. There's no way for me to get rid of the error, or to get RTC working again.
> 
> I tried this with users that have the `edit_with_others` capability, and those without. Both situations result in the exact same error. I've attached the react minified error.
> 
> CleanShot 2025-10-27 at 14 47 41 CleanShot 2025-10-27 at 14 56 11
> I've run a `distclean` on the Gutenberg branch, and cleaned out the RTC plugin's build directory as well. I've also tried to wipe away the vip dev-env and the same error comes up.
> 
> What I'm able to do though is collaborate on a post the normal way, which is to take over a post and see the lock show up on the other user.

Same for me

chriszarate:
Closing this in favor of a different approach: Privileged users can turn off collaborative editing for a single post using a toggle or button.

PR #78, opened 2025-10-24
Add notifications for users opening or closing a post

ingeniumed:
### Description

We now have notifications when a user opens and closes a post, similar to the content related toasts that we have in place.

The notifications have been broken into two:

- Collaborator joins a post
- Collaborator leaves a post

CleanShot 2025-10-24 at 15 12 30

I decided to group the two under notifications -> collaborator, with each toggle being labelled joins/leaves. This way, we can expand this to content related notifications and any other piece of change that we consider important.

CleanShot 2025-10-24 at 14 35 35

CleanShot 2025-10-24 at 14 35 50

### Testing

- Open a post, and ensure notifications are shown.
- Open the same post in another browser and ensure a notification is shown in the original browser only.
- Now leave the post in one of the browsers, and ensure a notification is shown in the other browser.


src/components/rtc-settings-panel.tsx
```diff
@@ -75,6 +88,14 @@ export function RTCSettingsPanel() {
 		setSelfAwarenessEnabled( enabled );
 	};
 
+	const handleToggleNotificationsForJoin = ( enabled: boolean ) => {
+		setNotificationsForJoinEnabled( enabled );
```
ingeniumed: My only concern with this approach is that, could we have too many settings to manage as we add more notifications? We are going to have two more toasts to manage so that'll be 4 notification settings to manage. I don't have a better solution that doesn't just do this exact same thing in a different manner.

src/components/rtc-settings-panel.tsx
```diff
@@ -75,6 +89,14 @@ export function RTCSettingsPanel() {
 		setSelfAwarenessEnabled( enabled );
 	};
 
+	const handleToggleNotificationsForCollaboratorJoining = ( enabled: boolean ) => {
```
ingeniumed: My only concern with this approach is that, could we have too many settings to manage as we add more notifications? We are going to have two more toasts to manage so that'll be 4 notification settings to manage. I don't have a better solution that doesn't just do this exact same thing in a different manner. This isn't meant to be a big concern right now, just me thinking out loud.

src/hooks/use-collaborator-activity-notifications.ts
```diff
@@ -0,0 +1,103 @@
...
+	UserState,
+	store as awarenessStore,
+} from '@/store/awareness-store';
+import { SettingsStoreSelectors } from '@/store/settings-store';
+
+/**
+ * Custom hook to show notifications when users join or leave the post.
+ * Notifications are shown based on the settings configured by the user.
+ */
+export function useCollaboratorActivityNotifications() {
```
ingeniumed: My logic for calling this hook `CollaboratorActivity` is that, this can be the single area where all notifications influenced by a collaborator's activity could go.

That's why the logic for actually sending notifications when a user joins/leaves is under presence notifications. It's a presence change related to collaborator activity.

pkevan:
Tested this, and one thing that stood out, was when you open the and close a new session in quick succession, the snackbar item doesn't necessarily react quick enough. I guess this could be down to the implementation of `initial` and `active` users which get merged (https://github.com/Automattic/vip-real-time-collaboration/pull/78/files#diff-a78a7b5c5932adbd78a743a17a070a2e0c6ffd0c276506e8cbff45e8710da8d3R97)?

to replicate:
- user 1 is editing post
- user 2 joins, and quickly leaves (they made a mistake or didn't mean to join)
- notification shows user 2 joining, but possibly due to timeout of item, the corresponding leave one doesn't appear until after the join notification has disappeared

pkevan:
> The notifications have been broken into two:
>
> Collaborator joins a post
> Collaborator leaves a post

Perhaps we should describe this in terms of connection to the post, i.e. `connected` `disconnected` or `enter` and `exit`, join/left feels somewhat ambiguous, particularly when `left` becomes `leave`.

chriszarate (changes requested):
Can you look into hooking into the `AwarenessManager` instead of this separate hook? We are already subscribing to changes there via `subscribeToUserChanges()`, and we have the advantage of aligning with existing presence indicators and debouncing that we're doing.

src/store/settings-store.ts
```diff
@@ -101,14 +131,22 @@ const selectors = {
 		const { isSelfAwarenessEnabled } = state;
 		return isSelfAwarenessEnabled;
 	},
+	isNotificationsForCollaboratorJoiningEnabled( state: SettingsState ) {
```
chriszarate: it's getting a bit messy in here, I think we should move to a single action / selector for all settings that accept an enum argument for the setting in question.

chriszarate:
> We are already subscribing to changes there via `subscribeToUserChanges()`

We can probably just dispatch these notifications from the existing action creators that manage user state (`upsertUser`, `removeUser`)

chriszarate:
Not a blocker or feedback on this PR, but I just want to point out that editing a post can also involve editing other entities (e.g., templates). We are positioning the post entity as the "parent" entity, which I think makes intuitive sense. Just pointing out that we may eventually want to represent a user's connection / presence on an entity-by-entity basis.

src/store/settings-store.ts
ingeniumed: Yip, that's one of the concerns I mentioned [here](https://github.com/Automattic/vip-real-time-collaboration/pull/78#discussion_r2458723697) albeit for a different entry point. I'll experiment with your suggestion, and hopefully that'll do the trick.

ingeniumed:
I had initially put these notifications in the awareness manager, but I felt like it worked better as a hook because it was a personalized notifications akin to the higlighting/cursor/other features that we have. If we do expand to other notifications, for say other entities it'd be useful to have this one notification hook that can be customized for each entity.

I'm not tied to having this as a hook or in the awareness manager personally, so I'm happy to try and move this to the awareness manager to see how that looks like.

ingeniumed:
> > We are already subscribing to changes there via `subscribeToUserChanges()`
> 
> We can probably just dispatch these notifications from the existing action creators that manage user state (`upsertUser`, `removeUser`)

I had experimented with this as well. Decided against it for the same reason as above. I wanted to try and set up some sort of an event bus that could handle notifications for all use cases down the line. We would just add to that single class and it'd work.

I'll experiment with this, to see if I could still keep that single generic notifications framework intact.

ingeniumed:
> Tested this, and one thing that stood out, was when you open the and close a new session in quick succession, the snackbar item doesn't necessarily react quick enough. I guess this could be down to the implementation of `initial` and `active` users which get merged (https://github.com/Automattic/vip-real-time-collaboration/pull/78/files#diff-a78a7b5c5932adbd78a743a17a070a2e0c6ffd0c276506e8cbff45e8710da8d3R97)?
> 
> to replicate:
> 
> * user 1 is editing post
> * user 2 joins, and quickly leaves (they made a mistake or didn't mean to join)
> * notification shows user 2 joining, but possibly due to timeout of item, the corresponding leave one doesn't appear until after the join notification has disappeared

Figured this out - forgot to add the clientID at the end of the id of the notification toast. I recall running into this for the workflow plugin so I remembered the cause.

src/store/settings-store.ts
ingeniumed: This has been completed in https://github.com/Automattic/vip-real-time-collaboration/pull/78/commits/3f51ac055d5bae2260b59dca973c628a930daec1

ingeniumed:
### Summary of new changes

- The verbiage has been changed to say a `{ User } has entered/exited a post'`. Thanks @pkevan for the feedback on that.
- The notifications have been moved to the awareness store, and I've extracted the notifications into a single class as much as possible. 
- There's enums now in place for notifications to identify each one. The value of each enum is the id used in the toast, and the clientID has been appended to ensure repeated toasts aren't considered to be duplicates.
- The settings store has been simplified down to a single enum based action/selector. I couldn't simplify the settings-panel more as the `UseSelect` needs to be the same on each render, given a state/parameters or else react warns about re-renders happening.
- Tests have been added. I had to skip adding tests for the toasts code as node's mock isn't as good as jest. We'll likely need to introduce a mock to do this, punting this to another PR.

CleanShot 2025-10-27 at 10 03 09

src/store/awareness-store.ts
```diff
@@ -116,6 +131,12 @@ const reducer = ( state = DEFAULT_STATE, action: AwarenessAction ): AwarenessSto
 					// No changes, don't update the state.
 					return state;
 				}
+			} else {
+				const content = getUserPresenceNotificationContent(
```
ingeniumed: Instead of trying to check for various conditions here, I've left the logic behind if a notification should be sent or not for the actual notification sending method.

src/utilities/notifications.ts
```diff
@@ -1,5 +1,21 @@
+import { store as noticesStore } from '@wordpress/notices';
+import { dispatch, select } from '@wordpress/data';
 import type { UserState } from '@/store/awareness-store';
+import { store as settingsStore, Setting } from '@/store/settings-store';
 
+export enum NotificationType {
+	PostRestored = 'remote-user-post-restored',
```
ingeniumed, line 9: I'm going to be making a follow up PR to actually enable these notifications, so I'll tweak the enums then. It's fine if they aren't 100% accurate because this code is not being used at the moment and will be changed by me soon.

src/utilities/notifications.ts
```diff
@@ -20,3 +43,75 @@ export function getPostUpdatedNotificationContent( userState: UserState, status:
...
+ */
+export function getUserPresenceNotificationContent(
+	userState: UserState,
+	type: NotificationType
+): string {
+	let action = type === NotificationType.UserEntered ? 'entered' : 'exited';
+	return `${ userState.name } has ${ action } the post.`;
+}
+
+function shouldSendNotification( userState: UserState, type: NotificationType ): boolean {
```
ingeniumed: This is the notification skipping code I've added. I think this is better suited to be here, so we can simplify the actual usage of these methods.

src/utilities/notifications.ts
```diff
@@ -20,3 +43,75 @@ export function getPostUpdatedNotificationContent( userState: UserState, status:
...
+/**
+ * Send a notification to the editor.
+ *
+ * Certain notifications can be skipped based on user settings, or scenarios.
+ *
+ * @param content the notification content
+ * @param userState the user state of the user related to the notification
+ * @param type the type of notification
+ */
+export function sendNotification(
```
ingeniumed, line 125: I didn't think it's critical if these notifications are sent or not, in the event that they are skipped. Hence why its void - treat it like a send it and forget it type of function. I don't think this needs to be a boolean right now, we can always revisit this later.

src/store/settings-store.ts
```diff
@@ -86,30 +60,21 @@ const reducer = (
...
-		const { isAwarenessCursorsEnabled } = state;
-		return isAwarenessCursorsEnabled;
-	},
-	isDebugToolsEnabled( state: SettingsState ) {
-		return isDevelopment() ? state.isDebugToolsEnabled : false;
-	},
-	isSelfAwarenessEnabled( state: SettingsState ) {
-		const { isSelfAwarenessEnabled } = state;
-		return isSelfAwarenessEnabled;
+	getSetting( state: SettingsState, setting: Setting ): boolean {
```
chriszarate, line 63: nice

src/store/settings-store.ts
```diff
@@ -6,36 +6,37 @@ import { loadFromLocalStorage, saveToLocalStorage } from '@/utilities/local-stor
 const STORE_NAME = 'vip-real-time-collaboration/settings';
 const LOCAL_STORAGE_KEY = 'vip-rtc-settings';
 
+export enum Setting {
+	AWARENESS_AVATARS = 'isAwarenessAvatarsEnabled',
+	AWARENESS_CURSORS = 'isAwarenessCursorsEnabled',
+	DEBUG_TOOLS = 'isDebugToolsEnabled',
+	SELF_AWARENESS = 'isSelfAwarenessEnabled',
+	USER_ENTER_NOTIFICATION = 'isUserEnterNotificationEnabled',
+	USER_EXIT_NOTIFICATION = 'isUserExitNotificationEnabled',
```
chriszarate: While string enum values are arbitrary, I find this naming a bit confusing since it mirrors the variable names when settings are selected. Generally it's a standard to mirror the member name:
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

It's minor, but makes it much better when scanning / searching quickly so that you aren't conflating the two (enums vs values).

src/utilities/notifications.ts
```diff
@@ -1,5 +1,23 @@
...
+	UserExited = 'remote-user-post-user-exited',
+}
+
+/**
+ * Get the content of a post restored notification.
+ *
+ * @param userInfo the user info of the user related to the notification
+ * @returns the content of the post restored notification
+ */
 export function getPostRestoredNotificationContent( userInfo: UserInfo ): string {
```
chriszarate: I think it would be much better if `sendNotification` controlled the notification content instead of allowing it to be passed in. It's bound to lead to messy logic scattered all over the codebase and inconsistent messaging. And already it would help with code duplication.

For the case where additional context is needed (e.g., `status`), there can be another export like `sendPostUpdatedNotification` (that wraps `sendNotification`).

ingeniumed: I had planned to do that in a follow up PR that'd enable the post restored/post updated notifications. It sorta aligns with what I mentioned [here](https://github.com/Automattic/vip-real-time-collaboration/pull/78#discussion_r2464280824).

Lemme give it another shot in this PR anyways.

src/store/settings-store.ts
ingeniumed: Good catch, that should have been a follow up fix from my end after I fixed all the hooks to use the new `get/set settings` way.

ingeniumed: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/78/commits/398f65245191700eaded307c2da28b7d3882dec8

src/utilities/notifications.ts
ingeniumed: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/78/commits/398f65245191700eaded307c2da28b7d3882dec8.

I decided to just make the one function to send the notification, rather than making another export. I'm going to making a follow up PR to actually enable these notifications and can revisit that then.

src/utilities/notifications.test.ts
```diff
@@ -18,59 +12,301 @@ const baseUserInfo: UserInfo = {
...
-		const result = getPostRestoredNotificationContent( baseUserInfo );
+describe( 'notifications', () => {
+	let sendNotification: typeof import('./notifications').sendNotification;
+	let notificationTypes: typeof import('./notifications').NotificationType;
+	let createNoticeMock: any = null;
+	let getSettingMock: any = null;
 
-		assert.strictEqual( result, 'Alice restored newer content from the server.' );
-	} );
+	before( async () => {
```
ingeniumed: Used Selection tests as inspiration for this. It did require some experimenting to get it exactly right

ingeniumed:
The tests have been updated finally as well.

src/components/rtc-settings-panel.tsx
```diff
@@ -103,36 +94,52 @@ export function RTCSettingsPanel() {
...
+						}
+					/>
+
+					<ToggleControl
+						label="Exits"
+						checked={ isUserExitNotificationEnabled }
+						onChange={ ( enabled: boolean ) =>
+							setSetting( Setting.USER_EXIT_NOTIFICATION, enabled )
+						}
+					/>
```
chriszarate, line 142: Do you think these should be a single setting? Why would someone want one and not the other?

src/utilities/notifications.test.ts
```diff
@@ -18,59 +12,301 @@ const baseUserInfo: UserInfo = {
...
+					};
+				},
+			},
+		} );
 
-		assert.strictEqual( result, 'Restored newer content from the server.' );
+		// @ts-expect-error: TS2702 Dynamic import -- used to navigate import issues mentioned above.
+		const notificationsModule = await import( './notifications' );
+		sendNotification = notificationsModule.sendNotification;
+		notificationTypes = notificationsModule.NotificationType;
```
chriszarate: o_O

What the hay? This is the code under test, why not just import it?

chriszarate: Ok, I read the docs, I get it now. Node.js mocks are wild. I cleaned things up and fixed type errors in 4bea2327db9860f0fcbb5cddee3f44b246bf33b8

chriszarate (approved):
Thanks for all the work here, gonna merge to make release

PR #82, opened 2025-10-27
Adding required styles

smithjw1:
Our telemetry system can't target empty divs, but if you add a minHeight, it works. Not sure why, but I think this is a benign style that will work for us.

The marginTop is to match the h3 above, in experimenting we need some sort of margin, I'm open to changes here.


chriszarate (changes requested):
Target this class in `rtc-overlay.scss` and add styles there, please.

Run linting with `npm run lint`.

smithjw1:
> Target this class in `rtc-overlay.scss` and add styles there, please.

@chriszarate I created a new SCSS file, which I thought was a bit safer. I know it's highly unlikely the overlay isn't there and the panel is, but if it happened, then the sidebar styles wouldn't be loaded.

I know adding another import adds some overhead, happy to hear it isn't worth it in this case.

src/components/rtc-settings-panel.tsx
```diff
@@ -31,6 +31,8 @@ const { unlock } = __dangerousOptInToUnstableAPIsOnlyForCoreModules(
 
 const { EditorPresence } = unlock( editorPrivateApis );
 
+import '@/components/rtc-settings-panel.scss';
```
chriszarate: Please put imports at the top of the file

chriszarate:
> I created a new SCSS file, which I thought was a bit safer.

That's fine, but in doing so please copy over all the style definitions used in this `RTCSettingsPanel` from `rtc-overlay.scss`. You can reference them by class name (`className`).

smithjw1:
Looking more closely at the code, there's no real danger to including the CSS in the overlay. It gets loaded from the panel.

chriszarate (approved):

PR #86, opened 2025-10-28
Add collaborators popover menu

shekharnwagh:
# Description

Adds a new collaborators popover that displays all active users in a clickable list which is opened on clicking the collaborators menu with avatars in the top editor bar. Users can click on any collaborator to scroll to their cursor position in the editor. The popover replaces the previous sidebar collaborators list section which has now been removed.

Added a cursor registry to store clientId and cursor ref mappings to facilitate scrolling to cursor positions. This uses a singleton pattern instead of React Context as context propagation was breaking between `EditorPresence` and `BlockCanvasCover.Fill`.

User emails require 'edit' context to be accessible via the REST API which required additional `getUser` call in addition to `getCurrentUser` which only returns the data with `view` context

## Screenshots

| Screenshot 2025-11-03 at 9 19 27 PM |
| :---: |
| _List of active collaborators, including names and email addresses opened on clicking menu button in top bar_ |

### Video

https://github.com/user-attachments/assets/9122fd24-eae8-4883-b4aa-40b5f51949f9


src/components/avatars.tsx
```diff
@@ -30,7 +37,11 @@ export function Avatars() {
 	const remainingUsersText = remainingUsers.map( ( { userInfo } ) => userInfo.name ).join( ', ' );
 
 	return visibleUsers.length > 1 ? (
-		<>
+		<div
+			className="vip-real-time-collaboration-avatars-container"
+			onMouseEnter={ () => setIsPopoverVisible( true ) }
+			onMouseLeave={ () => setIsPopoverVisible( false ) }
```
chriszarate: In order to be accessible, the popover must be triggerable from a keyboard-focusable element, typically a link, `<button>`, or element with `role="button"`.

src/utilities/entity.ts
```diff
@@ -1,18 +1,29 @@
...
+import { select, resolveSelect } from '@wordpress/data';
 
 import type { WordPressUserInfo } from '@/store/awareness-store';
+import type { User } from '@wordpress/core-data/build-types/entity-types';
+
+async function getUserEmail( userId: number ): Promise< string | undefined > {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+	const user = ( await resolveSelect( coreStore ).getUser( userId, {
+		context: 'edit',
+	} ) ) as User< 'edit' > | undefined;
```
chriszarate, line 11: Non-admins will not have permission for this request (for other users)

src/utilities/entity.ts
```diff
@@ -1,18 +1,29 @@
...
+	return user?.email;
+}
 
 export async function getCurrentUserInfo(): Promise< WordPressUserInfo > {
-	const { avatar_urls: avatarUrls, id, name, slug } = select( coreStore ).getCurrentUser() ?? {};
+	const currentUser = select( coreStore ).getCurrentUser() ?? {};
+	const { avatar_urls: avatarUrls, id, name, slug } = currentUser;
```
chriszarate, line 17: You should be able to destructure `email` from here. That way each user supplies their own info.

shekharnwagh: I tried that it was emtpy. The return type of getCurrentUser is `User< view >` which doesn't have `email` - https://github.com/WordPress/gutenberg/blob/c54fa0beb059a2e3b2d2f5a32f26ab47598be0b6/packages/core-data/src/entity-types/user.ts#L36-L39

src/utilities/entity.ts
shekharnwagh: We only call this for current user's ID, so wouldn't that be safe ?

chriszarate: Ah yeah, I thought this was being called from collaborator list. In that case we should just use this select instead of getCurrentUser.

shekharnwagh: This doesn't work with user id as `me`, so we need to fetch the current user id first with `getCurrentUser` and then use the id in next request.

chriszarate: Fun, carry on. I can't even find where the `getUser` selector is defined. 🫤

shekharnwagh: They dynamically generated here for entitites using [entitySelectors](https://github.com/WordPress/gutenberg/blob/c54fa0beb059a2e3b2d2f5a32f26ab47598be0b6/packages/core-data/src/index.js#L35-L58) method which is used when creating the store [here](https://github.com/WordPress/gutenberg/blob/c54fa0beb059a2e3b2d2f5a32f26ab47598be0b6/packages/core-data/src/index.js#L121).

Its also mentioned in the `core-data` package readme - https://github.com/WordPress/gutenberg/tree/c54fa0beb059a2e3b2d2f5a32f26ab47598be0b6/packages/core-data#plural

chriszarate: Yay, thanks for the pointers! Dynamically named functions, what a foil for my `ripgrep`ing. :)

src/components/collaborators-list.tsx
```diff
@@ -0,0 +1,102 @@
...
+		const success = cursorRegistry.scrollToCursor( clientId, {
+			behavior: 'smooth',
+			block: 'center',
+			highlightDuration: 2000,
+		} );
+
+		// Optionally close the popover after successful scroll
+		if ( success ) {
+			setIsPopoverVisible( false );
+		}
```
shekharnwagh: I am not entirely sure if we should close the popover when we click on a collaborator. Only added it to de-clutter it when we scroll, by happy to keep the popover open even after click.

maxschmeling: I could make an argument either way on this, so I'm fine letting it close for now.

src/contexts/cursor-registry-context.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+	/**
+	 * Remove all cursor elements from DOM and clear the registry.
+	 * Uses stored refs instead of DOM queries for better performance.
+	 */
+	public removeAll = (): void => {
+		// Remove each cursor element using stored refs
+		this.cursorMap.forEach( element => {
+			element.remove();
+		} );
```
maxschmeling:
```suggestion
		this.cursorMap.forEach( element => element.remove() );
```

maxschmeling (approved):
This is working pretty well.

We should improve the accessibility by allowing keyboard navigation through the list (up/down arrows), and as @shekharnwagh pointed out to me we'll probably want some feedback for users in the list that don't have a cursor position to scroll to. But this is a great first pass to get some feedback.

PR #87, opened 2025-10-28
Namescape custom metric names

chriszarate:


websocket-server/metrics.ts
```diff
@@ -31,40 +32,40 @@ const recentDisconnects = new Map< string, number >(); // connection_id -> disco
  * ------------------------------------------------------------
  */
 const connectedClientsGauge = new Gauge( {
-	name: 'websocket_connected_clients',
+	name: `${ METRICS_NAMESPACE }_connected_clients`,
```
pkevan, line 35: not necessarily related to this PR, but wondering if it would be helpful to have a histogram of connected clients, with buckets `[ 1, 5, 10, 20, 50, 100, 200, 500, 1000 ]`?

maxschmeling (approved):

websocket-server/metrics.ts
chriszarate: i don't know, what question would that answer that a gauge + functions cannot?

pkevan: was thinking for all websocket connections (across VIP) to gather an average/highest number - but if that's possible elsewhere then unnecessary. i.e. no-one hits 1000.

chriszarate: I think you can do `max_over_time(gauge_metric[time_period])` for that

PR #88, opened 2025-10-29
Update the post notifications

ingeniumed:
### Description

We can now send notifications in the editor when a post is updated. It accompanies [this PR](https://github.com/WordPress/gutenberg/pull/72851/files) that adds the necessary functions in place.

I've removed the post restored notifications. Instead, I've just added in the post being updated notifications. Once we observe this in practice, and get feedback we can revisit this decision. Technically, adding in the post restored notifications would be an extra feature/toast so this makes sense imo.

CleanShot 2025-10-29 at 10 22 56

CleanShot 2025-10-29 at 11 07 07


ingeniumed:
Note: I'm thinking of a follow up to move all the settings to a settings screen, so that's already on my mind.

src/components/rtc-settings-panel.tsx
```diff
@@ -121,6 +124,18 @@ export function RTCSettingsPanel() {
 						{ __( 'Notifications', 'vip-real-time-collaboration' ) }
 					</Heading>
 
+					<Heading level={ 2 } style={ { marginBottom: '2px' } }>
+						{ __( 'Post', 'vip-real-time-collaboration' ) }
+					</Heading>
+
+					<ToggleControl
+						label="Updates"
```
chriszarate: Is it an "update" or a "save / publish" event?

src/utilities/notifications.ts
```diff
@@ -92,6 +76,14 @@ function shouldSendNotification(
 		return false;
 	}
 
+	// If notifications for post updates is disabled, skip.
+	if (
+		type === NotificationType.PostUpdated &&
+		! select( settingsStore ).getSetting( Setting.POST_UPDATE_NOTIFICATION )
+	) {
+		return false;
+	}
```
chriszarate: This function could be DRYd up with a map of notification types to their governing setting.

src/components/rtc-settings-panel.tsx
ingeniumed: I called it update to keep it generic enough to mean both. It's fine to say save/publish as well.

src/utilities/notifications.ts
ingeniumed: Yip, that's fair it'd simplify it down even more.

src/components/rtc-settings-panel.tsx
ingeniumed: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/88/commits/33c7ced7642684972c698212600baf40b4609d19

src/utilities/notifications.ts
ingeniumed: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/88/commits/33c7ced7642684972c698212600baf40b4609d19

ingeniumed:
CleanShot 2025-10-29 at 13 48 13

pkevan:
> Note: I'm thinking of a follow up to move all the settings to a settings screen, so that's already on my mind.

This makes even more sense when considering that the settings can span multiple post types, which could be fine, but creates a natural expectation that they are distinct.

pkevan:
> CleanShot 2025-10-29 at 13 48 13

Is `Post` the correct labelling here?

ingeniumed:
> > CleanShot 2025-10-29 at 13 48 13
> 
> Is `Post` the correct labelling here?

I can't access that screenshot, but I was debating between content and post. I settled on post ultimately, though happy to change this. Personally, I think post is fine for now. If we are going to be moving this to a settings UI in another PR we can alter it then.

ingeniumed:
PR is targeting https://github.com/WordPress/gutenberg/pull/72851/files now.

chriszarate (approved):

PR #89, opened 2025-10-29
Enable eslint and fix linting issues

ingeniumed:
### Description

As noted internally, eslint wasn't running correctly as the path wasn't provided. Providing a path also means the actual eslint problems will start to come up so those have to be fixed.

Everything's been fixed, and the only folders that have been ignored are:

- yjs-inspector: This is the inspector code that we don't own
- tests/e2e: All of these e2e tests are currently disabled as they are flaky. There's no point trying to fix them if in trying to fix it, half the code is changed. I've punted it to when the tests are actually fixed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,7 @@ build
 node_modules
 vendor
 *.php
-websocket-server
 yjs-inspector
 tests/e2e
 examples/
+websocket-server
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,5 +4,4 @@ vendor
 *.php
 yjs-inspector
 tests/e2e
-examples/
 websocket-server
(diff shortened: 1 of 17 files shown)


pkevan:
There is a certain level of irony in a PR that fixes linting issues having linting issues 🙃

ingeniumed:
> There is a certain level of irony in a PR that fixes linting issues having linting issues 🙃

It's odd because I can't replicate this locally haha

chriszarate (approved):
thanks!

ingeniumed:
I've pushed up one more fix:

- Convert the lint staged config to a js based one, and run it as a function. This is from [this issue](https://github.com/lint-staged/lint-staged/issues/825) to workaround the fact that tsc doesn't use a tsconfig if files are provided. The function based call prevents that from happening just for lint staged. It leaves the normal commands alone.
- Simplify the eslintrc.js for websocket-server. I can specify true for the parserOption#project to pick up the nearest tsconfig.json, and as long as the dirname is provided it'll use the right one.

PR #90, opened 2025-10-29
Add telemetry logging UI

acicovic:
This is the first PR in a series of work related to telemetry, in order to keep PR size down and specific scope as much as possible.

This PR introduces:
1. A "Log telemetry using Logger" checkbox under Debug Tools, which is only available during development mode
2. A new option in the settings store to keep the value of the checkbox
3. A `setEnabled()` function in our Logger, allowing us to dynamically enable and disable log output

image

Functionality has been tested locally with other parts of the telemetry system.


src/utilities/logger.ts
```diff
@@ -20,8 +21,12 @@ export class Logger {
 		this.namespace = `vip-rtc${ localNamespace ? `:${ localNamespace }` : '' }`;
 	}
 
+	public setEnabled( enabled: boolean ): void {
+		this.enabled = enabled;
+	}
```
chriszarate, line 26: How would this be used and why? Can distracted developers just filter their console?

It's not called and can't be called from the browser console. Can we remove this until we have a defined use case?

chriszarate (approved):

src/utilities/logger.ts
acicovic: I had code outside of the PR that was using this, though this is not relevant at this point as I'll be closing this PR.

acicovic:
At the phase of development this was posted, I was logging a lot of things for debugging purposes. However, the number of logging is now much less and most of it is error-related. As such, I don't think we'll be needing this so I'm closing it.

PR #93, opened 2025-10-29
Update lint-staged config

ingeniumed:
### Description

Our lint-staged currently runs the entire lint command for all files types rather than targeting just the js/ts files. This change specifically runs only the `lint:js` command which internally has been updated to include the `check-types`. This also means the CI command can be tweaked to account for that.

Also, the issue that led to changes to our lint-staged config in #89 has been included here as well.


chriszarate (commented):
🆒 

can you update the required checks for `trunk` after merging?

PR #94, opened 2025-10-30
Move settings to a settings page

ingeniumed:
### Description

Our settings currently reside in the editor, and that's getting quite complicated as we add more settings. Moving those to a proper settings screen means we can clean up the editor, and make this screen a location for other useful items down the line.

This PR raises an important question:

Should the settings be user specific, or site specific? The JS code is agnostic of which option we pick, but the PHP code would change.

The PHP code uses options under the hood, which would need to be changed to something user specific if we want to store this on a per user basis.

Also, the code could be cleaned up a touch more so with all this in mind I've left this in draft status.

CleanShot 2025-10-30 at 15 11 09


inc/Settings/Settings.php
```diff
@@ -0,0 +1,288 @@
...
+					}
+				}
+			}
+		}
+
+		return $default_options;
+	}
+
+	public static function register_settings(): void {
+		register_setting( self::SETTINGS_PAGE_SLUG, self::OPTION_NAME );
```
chriszarate: Settings pages are not accessible to non-Administrators, so we can't put user-level settings there.

ingeniumed: As discussed internally, these settings would remain user specific. So I'll be updating this accordingly.

ingeniumed:
CleanShot 2025-11-05 at 14 26 25

The new settings page just has a single checkbox, that communicates what happens if it's disabled. This'll be a lot simpler than a button or anything else imo.

vip-real-time-collaboration.php
```diff
@@ -51,11 +52,23 @@
 		return;
 	}
 
+	// Initialize the core classes needed to run the barebones plugin, even if RTC is disabled.
+	new Settings();
+	// This needs to be intialized to account for manually enabling the experiment via the GB settings.
+	// That also needs to be reflected even if RTC is disabled.
+	new Compatibility();
```
ingeniumed, line 62: It is possible to manually enable the experiment without our plugin, which means when you disable the plugin the experiment is still enabled. So, this accounts for the case where the experiment is enabled but the plugin is enabled AND the plugin's functionality is meant to be disabled.

Safari seems to have a problem if this case is not accounted for, which causes it to not load the crypto module as it messes up the Yjs module. I think it's related to our plugin, but I couldn't quite fix it. What I did realize in trying to fix this is that, we should be disabling the experiment if our plugin's settings says that RTC should be disabled.

inc/Settings/Settings.php
```diff
@@ -0,0 +1,151 @@
...
+
+		/** @psalm-suppress InvalidArgument */ // WordPress Settings API allows custom args.
+		add_settings_field(
+			'enable-vip-rtc',
+			__( 'Enable VIP Real-Time Collaboration', 'vip-real-time-collaboration' ),
+			[ __CLASS__, 'display_settings_checkbox' ],
+			self::SETTINGS_PAGE_SLUG,
+			'plugin-settings',
+			[
+				'label' => __( 'Disabling this, will disable the VIP Real-Time Collaboration plugin. This can be used in case of an emergency.', 'vip-real-time-collaboration' ),
```
ingeniumed: Happy to tweak this verbiage

pkevan: Perhaps I'm misunderstanding, but are we shifting the purpose so that now you need to check this for the plugin to work - if so how will that link with the integration center?

It might be less confusing to label this a emergency stop setting, rather than a setting which you have to turn off in an emergency (aka Don't click this if you want to stop 😂 )

Also, a radio button is probably preferable here (enable/disable, on/off) so that the action is clear.

ingeniumed: The idea is that, this gives you a quick way to disable collaboration without having to worry about disabling the entire integration. I can see why this would be confusing, which is why I tried to add a label to make it clear.

checked = enabled
unchecked = disabled

I can try to play around with a radio button or a toggle and see if that'd be better.

ingeniumed: The radio button works a lot better than the checkbox, as I was able to change the text to be a lot less confusing. Check it out, and let me know what you think.

ingeniumed:
CleanShot 2025-11-06 at 13 44 35

After feedback from @pkevan, I've changed the checkbox to a radio button and the verbiage has been simplified as well.

inc/Settings/Settings.php
```diff
@@ -0,0 +1,164 @@
...
+			'enable-vip-rtc',
+			__( 'Real-Time Collaboration', 'vip-real-time-collaboration' ),
+			[ __CLASS__, 'display_settings_radio' ],
+			self::SETTINGS_PAGE_SLUG,
+			'plugin-settings',
+			[
+				'field_id' => 'enable-vip-rtc',
+				'enabled_label' => __( 'Enable real-time collaboration', 'vip-real-time-collaboration' ),
+				'disabled_label' => __( 'Disable real-time collaboration', 'vip-real-time-collaboration' ),
+				'description' => __( 'Disable real-time collaboration functionaliy in case of an emergency.', 'vip-real-time-collaboration' ),
```
pkevan:
```suggestion
				'description' => __( 'Disable real-time collaboration functionality in case of an emergency.', 'vip-real-time-collaboration' ),
```

fixes spelling error

pkevan:
```suggestion
				'description' => __( 'Disable functionality in case of an emergency.', 'vip-real-time-collaboration' ),
```

Removes additional `real-time collaboration` string - we repeat the word four times in quick succession so I'd imagine users know what we are talking about at this point.

Other option is to label the buttons `Enable` `Disable`

ingeniumed: I wanted to mention real time collaboration as much as possible so that the consequence of this setting are known. I'm happy to drop the additional string from the description, but I'd want to keep it in in the radio button labels.

vip-real-time-collaboration.php
```diff
@@ -51,11 +52,23 @@
 		return;
 	}
 
+	// Initialize the core classes needed to run the barebones plugin, even if RTC is disabled.
+	new Settings();
+	// This needs to be intialized to account for manually enabling the experiment via the GB settings.
+	// That also needs to be reflected even if RTC is disabled.
```
chriszarate: Could you improve this comment a bit? I can't really follow it. Or just remove it.

vip-real-time-collaboration.php
```diff
@@ -51,11 +52,23 @@
 		return;
 	}
 
+	// Initialize the core classes needed to run the barebones plugin, even if RTC is disabled.
```
chriszarate: Perhaps a bit clearer?
```suggestion
	// Initialize the core classes needed to manage the Gutenberg experiment.
```

inc/Settings/Settings.php
chriszarate: FWIW I agree with Paul. There are better ways to communicate the importance of this setting.

chriszarate (approved):

vip-real-time-collaboration.php
ingeniumed: I can improve it. I'd like to leave it in so its clear why we have that code in place.

PR #102, opened 2025-11-04
Pass cursor registry as shared prop

chriszarate:
Previously, the `CursorRegistry` was positioned as React context (which it wasn't) and instantiated in the global scope. This global non-reactive state nearly always leads to issues down the road. 

Ideal reactive options are elusive:

- Refactor our cursor rendering as pure functional React components (complex).
- Wrap the entire block editor component tree in a real context provider (not possible?).
- Store pointers to cursor elements in the awareness store (moderately complex).

For now, we instantiate the `CursorRegistry` in the render loop using a ref to maintain references and describe the issue in a comment.

### DIFF
--- a/src/components/avatars.tsx
+++ b/src/components/avatars.tsx
@@ -4,14 +4,19 @@ import { useState } from '@wordpress/element';
 import { Avatar } from '@/components/avatar';
 import { CollaboratorsList } from '@/components/collaborators-list';
 import { useSortedAwarenessUsers } from '@/hooks/use-sorted-awareness-users';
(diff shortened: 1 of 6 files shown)


shekharnwagh (approved):

chriszarate:
Improved comments in https://github.com/Automattic/vip-real-time-collaboration/pull/102/commits/203ded699bff741dd4a87003fe5fcf81bbd93fbd. Thanks to @shekharnwagh for the conceptual walkthrough and understanding of the `ref` approach.

PR #103, opened 2025-11-05
Add collaborative session statistics

acicovic:
### Summary

This PR implements telemetry infrastructure to gather and log statistics for multi-user real-time collaborative editing sessions.

### Key Changes

**New Telemetry Infrastructure:**

- Implements `TelemetryLogger` base class providing a standard interface for logging telemetry data to multiple destinations (Logger, Pendo)
- Implements modular telemetry system under `src/telemetry/` for extensibility

**Session Statistics Tracking:**

- Implements `SessionStats` class to manage multi-user session data using Yjs shared state
- Implements session-specific logger (`SessionStatsTelemetryLogger`) that gathers multi-user session statistics
- Sessions are only tracked when 2+ users are present, and logged when 1 user remains

**Session Activity Management:**

- Implements `SessionActivityManager` to coordinate session lifecycle events
- Monitors user joins/leaves and transitions between single/multi-user states
- Detects document changes and marks locally active users
- Implements `UserInactivityTimer` with 30-minute inactivity timeout (no document changes) to detect idle sessions and trigger session logging
- Uses debouncing and throttling to avoid redundant updates

**Statistics Collected:**

- Post ID
- Count of maximum users during the session
- Count of active users (performed edits) vs. inactive users (view-only)
- Session duration and activity timestamps

**Integration:**

- Integrates `SessionActivityManager` initialization into `AwarenessManager` lifecycle
- Statistics stored in Yjs document under `session-stats` origin map
- Provides utilities for checking user connections and exporting session stats

### Technical Details

**Data Integrity and Safeguards:**

- Transaction-based updates to prevent race conditions
- Graceful degradation when telemetry data cannot be prepared
- Prevents duplicate initialization and handles reconnection edge cases
- Prevents race conditions when logging stats, with checks on user activity and session state
- Uses grace periods and validation to avoid duplicate or incomplete logs during quick user connect/disconnect cycles
- Uses leader election pattern to prevent duplicate logging from multiple clients
- Elects a fallback leader when the current leader disconnects

### Testing

- Basic functionality tested across the latest version of Brave, Vivaldi, Safari and Firefox.
- To test this, simulate a multi-user session locally. You will see output in the console when 2+ users join (session initialization) and when only 1 user remains in the session (final session stats).

Summary of all tested scenarios:

- Single user sessions (no logging occurs)
- Leader logs when non-leaders leave
- Non-leader takes over and logs when leader leaves
- Inactivity timeout triggers logging with correct stats
- Activity resumption starts new session
- Multiple inactivity/activity cycles log correctly
- Multiple tabs per user handled correctly
- Computer sleep/wake cycles
- Short disconnects and reconnections
- Browser back/forward navigation

### Future Work

Pendo integration for telemetry logging will be implemented in a follow-up PR. This will involve creating a WordPress REST API endpoint that receives the telemetry data and forwards it to the VIP Telemetry class.


src/telemetry/session-stats/SessionActivityController.ts
```diff
@@ -0,0 +1,402 @@
...
+	public static initialize( awareness: Awareness ): void {
+		if ( SessionActivityController.__instance ) {
+			if ( SessionActivityController.__instance.awareness?.doc !== awareness.doc ) {
+				SessionActivityController.destroySingletonInstance();
+			} else {
+				return;
+			}
+		}
+
+		if ( ! awareness?.doc ) {
```
ingeniumed: Should this be checked right away instead?

src/telemetry/session-stats/SessionActivityController.ts
```diff
@@ -0,0 +1,402 @@
...
+
+/**
+ * Coordinates collaborative session events and the execution of operations in
+ * response to them.
+ *
+ * Subscribes to session events, manages session state transitions, tracks user
+ * activity, responds to user and document changes, handles inactivity, logs
+ * session data.
+ */
+export class SessionActivityController {
```
ingeniumed: Nitpick: Feel like calling it a manager is better imo rather than a controller. A controller to me implies rest endpoints that'll be actively called.

Unless there's work planned in a follow up PR to add that in? I assumed the follow up work was to add it on the PHP side.

src/telemetry/session-stats/SessionActivityController.ts
```diff
@@ -0,0 +1,402 @@
...
+	 *
+	 * If an instance already exists with a different Yjs document, it is
+	 * destroyed first.
+	 *
+	 * @param awareness The Yjs Awareness instance to use
+	 */
+	public static initialize( awareness: Awareness ): void {
+		if ( SessionActivityController.__instance ) {
+			if ( SessionActivityController.__instance.awareness?.doc !== awareness.doc ) {
+				SessionActivityController.destroySingletonInstance();
```
ingeniumed: What would an example be of this being called? Awareness Manager is initialized once per client, and then errors subsequently.

This feels like a singleton anti-pattern, if we are re-creating the entire singleton instance when the awareness doc changes? Might be a naming thing, but I wouldn't expect a singleton to be destroyed like this if we are using this pattern unless absolutely necessary?

src/telemetry/session-stats/SessionActivityManager.ts
```diff
@@ -0,0 +1,402 @@
...
+	constructor( private awareness: Awareness ) {
+		this.sessionStats = new SessionStats( awareness, this.logger );
+
+		this.inactivityTimer = new UserInactivityTimer(
+			SESSION_TIMEOUT_IN_SEC * 1000,
+			() => this.sessionStats.getLastActivityTime(),
+			() => this.handleInactivityTimeout()
+		);
+
+		this.subscribeToUserChanges();
```
ingeniumed, line 67: Looking at this, I wonder if the class could be re-done such that:

- Instantiate it within the constructor of the awareness manager, so their lifecycles match. This means a single clientID at a time would have this class instantiated.
- Within the existing subscriptions, call in the methods on this class that require those subscriptions. That way, there's one single grouped awareness subscription for the awareness changes to be transmitted to. That was a performance optimization that had been mentioned on some forum posts (don't have the link on hand but I remember running into that as an improvement as more clients join).

src/telemetry/session-stats/SessionActivityController.ts
```diff
@@ -0,0 +1,402 @@
...
+			this.inactivityTimer.restart();
+			this.sessionStats.updateLastActivityTime();
+
+			// If this is a local change, mark the originating user as active.
+			if ( transaction.local ) {
+				const clientId = this.awareness.clientID;
+
+				try {
+					const { getActiveUsers } = select( awarenessStore );
+					const userId = getActiveUsers().get( clientId )?.userInfo?.id ?? null;
```
ingeniumed: You can also look for the `isMe` property, and that'll do the trick.

src/telemetry/session-stats/SessionStats.ts
```diff
@@ -0,0 +1,401 @@
...
+	private lastInitializationTime = 0;
+
+	/**
+	 * Initializes the class.
+	 *
+	 * @param awareness The Yjs Awareness instance to use for shared state
+	 * @param logger The Logger instance to use for logging
+	 */
+	constructor( private awareness: Awareness, private logger: Logger ) {
+		this.sessionMap = this.awareness.doc.getMap( SESSION_STATS_ORIGIN );
```
ingeniumed, line 53: Would check `awareness?.doc` before accessing the `getMap` on it out of safety 🤔

src/telemetry/session-stats/SessionStats.ts
```diff
@@ -0,0 +1,401 @@
...
+
+		const durationMs = Date.now() - sessionStartTime;
+
+		return Math.floor( durationMs / 1000 );
+	}
+
+	/**
+	 * Returns the current post ID.
+	 */
+	private getCurrentPostId(): PostId {
```
ingeniumed, line 150: Why do we wrap an existing selector offered by Gutenberg?

src/telemetry/session-stats/SessionStats.ts
```diff
@@ -0,0 +1,401 @@
...
+/**
+ * Type-safe keys of SessionStatsSchema.
+ */
+type SessionStatsKey = keyof SessionStatsSchema;
+
+/**
+ * Manages multi-user session statistics using Yjs Awareness for shared state.
+ */
+export class SessionStats {
+	private sessionMap: Y.Map< unknown >;
```
ingeniumed, line 40: This is different than how we are using other maps that we are setting on the doc. 

Usually, what we do this within a function rather than at the class level:

- Use observe or observeDeep to monitor for changes, if that's relevant
- Use transact to make changes to a map on the doc itself
- Store the doc in memory like in Manager, and access what's needed on that

Could this way have any issues related to state?

src/telemetry/session-stats/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Extracts unique user IDs from connected users in the awareness store.
+ *
+ * Filters out disconnected users and returns only valid positive integer user IDs.
+ *
+ * @returns Array of unique connected user IDs
+ */
+export function getConnectedUserIds(): number[] {
```
ingeniumed, line 42: Would it be worth adding methods that you think are useful, on the awareness store itself so they could be re-used elsewhere?

src/awareness-manager.ts
```diff
@@ -68,6 +69,7 @@ export class AwarenessManager {
 		}
 
 		AwarenessManager.__instance = new AwarenessManager( awareness, await getCurrentUserInfo() );
+		SessionActivityController.initialize( awareness );
```
ingeniumed: This feels like an odd place to initialize it imo. I feel like it should be in the awareness manager constructor in some way? This is assuming the other comment I made about changing how SessionActivityController is initialized.

src/telemetry/session-stats/SessionActivityController.ts
acicovic: I've added an `isMe` validation check to the user lookup. I kept the direct clientID map access for O(1) performance (this runs on every local document change), but now we validate with isMe before proceeding.

src/telemetry/session-stats/SessionStats.ts
acicovic: `awareness.doc` is technically guaranteed by the Awareness type definition (it's a required constructor parameter and non-optional property). Although I'm unsure where this could go wrong, I've added a defensive check.

From what I can see, we generally use `awareness.doc` across the codebase with the only exception being `convertRelativePositionToAbsolutePosition()` which can be called externally.

src/telemetry/session-stats/SessionStats.ts
acicovic: This wrapper is used in `exportStats()`, allowing partial success. If `getCurrentPostId()` fails, we still log the session with `postId: null` rather than losing all telemetry data. Without it, a selector error would trigger the outer catch and discard the entire export.

src/telemetry/session-stats/SessionStats.ts
acicovic: I don't think we could have an issue here. `doc.getMap()` is idempotent and returns a live reference, not a snapshot. Storing it should be safe and avoids repeated lookups. To put this in more words:

- `doc.getMap('key')` returns the same instance every time
- The map reference remains valid for the lifetime of the doc
- Changes to the map are reflected immediately in the stored reference
- The map is the actual shared state object, not a snapshot

Let me know if you think I could be missing something.

src/telemetry/session-stats/utils.ts
acicovic: In the context of this PR, I've tried to touch "outside files" the least amount possible. If any of these methods become useful for other parts of the code, we could definitely move them where it makes more sense. I don't know if this holds true currently.

acicovic:
@ingeniumed, thank you for your review.

I've addressed some of your feedback in https://github.com/Automattic/vip-real-time-collaboration/pull/103/commits/3505352a36efede7daa6c9966e6b7e49a44da701 and left comments in the remaining items. Let's keep the discussion going where needed, and let me know if the changes made look good to you or if you have more feedback.

Thanks!

chriszarate:
@acicovic This PR is pretty big! 

> Statistics stored in Yjs document under session-stats origin map

Could you explain why storing the stats in a Yjs doc is useful / necessary? If it's to aggregate stats among peers, could we instead get these answers using aggregation functions when querying the "snapshot-style" events?

acicovic:
> @acicovic This PR is pretty big!

Yeah, I'm sorry about this.

> Could you explain why storing the stats in a Yjs doc is useful / necessary?

If you have any particular concern about this implementation, let me know so I can look into it.

The Yjs doc enables us to log one accurate, fully-aggregated event per session.

- Shared session state: All clients need to track the same user list (`allUserIds`, `activeUserIds`) and timing data (`sessionTimeStart`, `sessionTimeLastActivity`) as users join/leave and make edits. Without this shared state, each client would only see its own partial view.
- Leader election: The `sessionInitializerClientId` stored in the Yjs doc enables leader election so exactly one client logs the final aggregated stats when the session ends. Otherwise, multiple clients would each log conflicting "session ended" events.
- Consistent inactivity detection: All clients check the shared `sessionTimeLastActivity` to detect the 30-minute timeout. Without shared state, clients would have different timeout values and log prematurely.
- Edge case handling: Clients joining mid-session need to see who was there before them. The shared Yjs doc makes this possible.

Potential issues with snapshot events + aggregation:

- No way to coordinate which client should log the final event
- Race conditions from multiple clients sending conflicting session-end events
- Missing data (users who left before you joined)
- Reconciliation logic required

acicovic:
Hi, I've pushed some additional commits to this PR. I'm not planning any additional changes, so this is ready for review.

src/websocket-client.ts
```diff
@@ -227,13 +227,22 @@ export function createWebSocketConnection( serverUrl: string ): ProviderCreator
...
+
 			if ( awareness ) {
+				// Wait for initial sync before initializing awareness.
+				await new Promise< void >( resolve => {
+					provider.once( 'sync', ( isSynced: boolean ) => {
+						if ( isSynced ) {
+							resolve();
+						}
+					} );
+				} );
```
chriszarate, line 240: Why do we need to await this sync event firing?

This feels like a potential bug. What if the `sync` event doesn't fire?

chriszarate:
> Potential issues with snapshot events + aggregation:
>
> - No way to coordinate which client should log the final event
> - Race conditions from multiple clients sending conflicting session-end events
> - Missing data (users who left before you joined)
> - Reconciliation logic required

I don't know what we are trying to measure, but I can't help thinking that this client-side approach is not a good conceptual match for tracking session length. Wouldn't it be much simpler to dispatch these events from the websocket server, which can easily track connected users and "inactivity" based on received messages?

I am very wary of storing statistics in the Yjs doc, since this can only grow and become more complex. In addition, the dependence between the maps risks an infinite loop.

src/telemetry/session-stats/SessionActivityManager.ts
```diff
@@ -0,0 +1,472 @@
...
+		if ( this.awarenessChangeHandler ) {
+			this.awareness.off( 'change', this.awarenessChangeHandler );
+		}
+
+		if ( this.documentUpdateHandler ) {
+			this.awareness.doc.off( 'update', this.documentUpdateHandler );
+		}
+
+		if ( this.sessionStatsObserver ) {
+			const sessionStatsMap = this.awareness.doc.getMap( SESSION_STATS_ORIGIN );
```
chriszarate, line 468: This isn't an origin

src/telemetry/session-stats/SessionActivityManager.ts
```diff
@@ -0,0 +1,472 @@
...
+	 */
+	private transactionHasContentChanges( transaction: Y.Transaction ): boolean {
+		if ( 0 === transaction.changed.size ) {
+			return false;
+		}
+
+		const contentMap = this.awareness.doc.getMap( RECORD_KEY );
+
+		// Maps that should not contain any content changes.
+		const excludedMaps = new Set( [
```
chriszarate, line 293: This feels likely to lead to future bugs

acicovic:
> Wouldn't it be much simpler to dispatch these events from the websocket server, which can easily track connected users and "inactivity" based on received messages?

I agree that this sounds like a simpler approach. The advantage of the approach proposed in this PR is that since we're hooking into the RTC session itself, it gives us the ability to measure much more, which is something that might come up.

> I am very wary of storing statistics in the Yjs doc, since this can only grow and become more complex. In addition, the dependence between the maps risks an infinite loop.

I'll be looking to understand the infinite loop possibility. About the first part, are there any viable alternatives you could suggest? For example, could we have another document? I don't see this working well with REST API endpoints writing the stats to the WP database. Also the stats data we're storing is both very small and simple, though I'm not familiar with the Yjs implications.

chriszarate:
> it gives us the ability to measure much more, which is something that might come up

I've mentioned the value of working backwards from the questions we need to answer, so that we can define what needs to be measured. Potentially over-engineering to maximize flexibility can lead to unwanted complexity. 

> About the first part, are there any viable alternatives you could suggest?

I would suggest that telemetry events be shipped by the websocket server, which has access to the clients and document, and controls the lifecycle of the session. I haven't verified this but it should also have access to the user info via the awareness document.

> I don't see this working well with REST API endpoints writing the stats to the WP database. 

Not sure I understand this.

acicovic:
> I don't see this working well with REST API endpoints writing the stats to the WP database.
> Not sure I understand this.

I was alluding to writing the stats into the database instead of the Yjs doc since you raised that concern.

smithjw1:
> I've mentioned the value of working backwards from the questions we need to answer, so that we can define what needs to be measured. Potentially over-engineering to maximize flexibility can lead to unwanted complexity.

I'm sympathetic to this line of thought as well. Here's the question I need answered today:

1. How many people are using real-time collaboration?
2. What's the average number of users in a real-time collaboration session?
3. How long do sessions last?
4. Which customers use it more than others?
5. Which customers have lots of editors?

In the future, I could see asking:

1. How many block errors are thrown in a session (If we hear reports that this is happening).
2. Were there mode transitions during the session?
3. What action caused an enforced mode transition?
4. Did the session hit per-document limits?
5. Did the transport layer return errors?
6. What was the average round-trip for the transport layer (more useful when we have more layers)?


> I would suggest that telemetry events be shipped by the websocket server, which has access to the clients and document, and controls the lifecycle of the session. I haven't verified this, but it should also have access to the user info via the awareness document.

If the WebSockets server can get us many of the additional questions, I'm all for it. I do wonder if that server will have all the information necessary to send a Pendo event with the correct parameters to track it back to the user and the client. If not right now, is it sensible to add that to another service?

chriszarate:
Thanks @smithjw1 for the detail!

> I'm sympathetic to this line of thought as well. Here's the question I need answered today:
>
> 1. How many people are using real-time collaboration?
> 2. What's the average number of users in a real-time collaboration session?
> 3. How long do sessions last?
> 4. Which customers use it more than others?
> 5. Which customers have lots of editors?

The data needed to answer these questions are straightforward to collect and ship from the websocket server. 

> In the future, I could see asking:

The answers to these questions are more nuanced, so I'll answer them separately.

> 1. How many block errors are thrown in a session (If we hear reports that this is happening).

This can probably only be observed on the client, but I'm not sure we can restrict our observation to errors that are specifically attributable to this plugin.

> 2. Were there mode transitions during the session?
> 3. What action caused an enforced mode transition?
> 4. Did the session hit per-document limits?

Data for these questions can be collected from the websocket server.

> 5. Did the transport layer return errors?

If we only care about errors that the transport layer actually returned, then we can collect this data from the websocket server. But other errors can only be observed on the client, like an error that results from a WiFi disruption. I'm not sure if we want these errors collected together.

> 6. What was the average round-trip for the transport layer (more useful when we have more layers)?

Assuming this is a measure of message latency, this is probably best observed by the client. Once we're ready to measure this, I might recommend a threshold or histogram approach.

In summary, I'm sure we will have some need for telemetry on the client in the future, but I think we can collect most of our data more easily from the websocket server. In doing so, we can also avoid the more complex parts of this PR (namely, the leader election and CRDT-syncing of telemetric data).

smithjw1:
@chriszarate thanks for that helpful response. I'm comfortable moving forward with the WebSockets implementation.

I assume that means the server will need to be able to send events to Pendo. It's fine if those events appear to come from the WordPress application. We'd use the same secret to send server-side events. Is there anything else necessary from the Pendo side?

chriszarate:
> Is there anything else necessary from the Pendo side?

I don't think so. I'd only note [our inability to get the user email](https://github.com/Automattic/vip-real-time-collaboration/pull/128) on the client side, which will also be true on the websocket server. Therefore, we will not be able to generate consistent user IDs.

smithjw1:
> Therefore, we will not be able to generate consistent user IDs.

I'm connecting some dots. The fix for #128 was necessary because we can't be sure a user has access to other users' email addresses. But all users can see the email address associated with their account.

Adding that to the awareness document doesn't serve a purpose outside of this telemetry work, but is it possible? Do you think it's a good idea? We could pass the hashed email that Pendo uses instead, so the WebSockets server doesn't have to do that work.

If we don't, these events will generate new, duplicate users and inflate our billable metric. If we can have a consistent user ID, it would be better, but I am also very keen to get visibility into usage at the account level in Pendo, so if it is a significant lift, we can go without.

chriszarate:
> We could pass the hashed email that Pendo uses instead, so the WebSockets server doesn't have to do that work.

I think this is doable and a good path forward. Assume we will provide consistent user IDs this way.

acicovic:
Closing this since we'll be changing our approach for stats.

PR #104, opened 2025-11-05
Support for collection syncing

chriszarate:
Support for collection syncing. Companion to https://github.com/WordPress/gutenberg/pull/73027

Allows for a connection representing a entity collection, with the magic string `collection` standing in for the object ID.

### DIFF
--- a/inc/Auth/SyncPermissions.php
+++ b/inc/Auth/SyncPermissions.php
@@ -47,7 +47,7 @@ final class SyncPermissions {
 		[ $entity_kind, $entity_name ] = $parts;
 
 		// Handle post type entities
-		if ( 'postType' === $entity_kind ) {
+		if ( 'postType' === $entity_kind && 'collection' !== $sync_object_id ) {
 			/**
 			 * For post entities, we only need the sync object ID (post ID) for permission checking.
 			 * The entity_name is Gutenberg's entity name which maps to post type name instead of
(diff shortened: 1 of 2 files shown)


inc/Auth/SyncPermissions.php
```diff
@@ -47,7 +47,7 @@ public static function can_sync(
 		[ $entity_kind, $entity_name ] = $parts;
 
 		// Handle post type entities
```
ingeniumed: This comment should be updated to note that we now support collections, and that so collections are also being excluded here.

src/websocket-client.ts
```diff
@@ -176,13 +176,16 @@ export function createWebSocketConnection( serverUrl: string ): ProviderCreator
...
-		objectId: ObjectID,
+		objectId: ObjectID | null,
 		doc: Y.Doc,
 		awareness?: Awareness
 	) {
 		try {
-			// For now, we only support traditional post types.
-			if ( ! objectType.startsWith( 'postType/' ) || ! parseInt( objectId, 10 ) ) {
+			// For now, we only support collections and traditional post types.
+			if (
```
ingeniumed, line 185: I'm assuming we are potentially expecting a case where the objectId could be 0, which is why you didn't go with the `if (objectId)`?

In that case, would it be better to drop the strict equality so undefined could be caught as well? Right now, only null is caught.

chriszarate: An explicit `null` objectId represents a collection. An `undefined` objectId is caught by `! parseInt( undefined, 10 )`.

This is more of a generic type guard since the calling code is in Gutenberg.

ingeniumed (approved):

PR #106, opened 2025-11-10
Add Pendo logging for session statistics

acicovic:
This is a work in progress and untested against Pendo at this time.


acicovic:
Hello,

I've closed #103 and #114 and I'm wondering if this should be closed too, or if it could be at least partly useful for the new implementation.

PR #107, opened 2025-11-10
Skip unnecessary user entered notification

ingeniumed:
### Description

Currently, if a user enters an existing collaboration session they get a redundant notification for every collaborator already in the session. They don't need to know that, and that's a redundant notification. This changes that, by skipping notifying them of all the users already in the session.

### Why

This happens because Yjs informs the new client of every new client that's been added from the existing client pool. There isn't a "entered At" field for the user unfortunately. 

### How

This is fixed by adding a timestamp to every user that enters a collaboration session, in the order in which they joined. This timestamp is then compared with the current "me" user to figure out who should get the notification and who shouldn't.

This also works for page refreshes now.

### Testing

 - Open a post in a browser
 - In another browser, open the same post
 - The only notification should be in the first browser, of the user in the second browser joining.
 - Refresh either page, and only 1 notification should show up

### DIFF
--- a/src/store/awareness-store.ts
+++ b/src/store/awareness-store.ts
@@ -22,6 +22,7 @@ export interface UserInfo extends WordPressUserInfo {
 	color: string;
 	isConnected: boolean;
 	isMe: boolean;
+	enteredAt: number;
 }
 
 export interface EditorState {
@@ -166,7 +167,16 @@ const reducer = ( state = DEFAULT_STATE, action: AwarenessAction ): AwarenessSto
 				}
 			} else {
 				const { userInfo } = action.payload.userState;
-				sendNotification( NotificationType.UserEntered, userInfo );
+				const currentMeUser = Array.from( state.userMap.values() ).find(
+					user => user.userInfo.isMe && user.userInfo.isConnected
+				);
+
+				sendNotification(
+					NotificationType.UserEntered,
+					userInfo,
+					undefined,
+					currentMeUser?.userInfo
+				);
 			}
 
 			state.userMap.set( action.payload.clientId, action.payload.userState );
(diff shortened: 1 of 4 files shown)


src/store/awareness-store.ts
```diff
@@ -166,7 +167,16 @@ const reducer = ( state = DEFAULT_STATE, action: AwarenessAction ): AwarenessSto
 				}
 			} else {
 				const { userInfo } = action.payload.userState;
-				sendNotification( NotificationType.UserEntered, userInfo );
+				const currentMeUser = Array.from( state.userMap.values() ).find(
+					user => user.userInfo.isMe && user.userInfo.isConnected
+				);
+
+				sendNotification(
```
ingeniumed, line 174: I had originally tried to keep the checks in here, but we want to keep all the notification sending logic in one place which is why it's been moved inside here. 

The function is a bit messy, so next time we add anything we may want to create separate functions perhaps for each notification type.

src/utilities/notifications.ts
```diff
@@ -101,29 +111,31 @@ function getContentForNotificationType(
...
+ * @param type The type of notification to send.
+ * @param userInfoToSendAbout The user info of the user related to the notification.
+ * @param status The status of the post (only relevant for PostUpdated notifications).
+ * @param currentMeUserInfo The user info of the current user (only relevant for user entered notifications).
  */
 export function sendNotification(
 	type: NotificationType,
-	userInfo: UserInfo,
-	status?: string
+	userInfoToSendAbout: UserInfo,
```
ingeniumed, line 121: Changed the name so it's not confusing

alecgeatches (approved):
Tested locally and works great!

PR #108, opened 2025-11-12
Add collaboration mode picker for view and edit modes

shekharnwagh:
## Description

Adds a new collaboration mode picker component in the editor toolbar that lets users switch between View mode and Edit mode.

Currently its just an UI without any functionality.

Depends on gutenberg PR - https://github.com/WordPress/gutenberg/pull/73196

## Screenshots

| Screenshot 2025-11-13 at 12 48 38 AM |
| :----: |
| Collaboration mode picker expanded |

### Screen Recording

https://github.com/user-attachments/assets/6f471770-44dd-4fb4-ac88-0bdd4029d707

### DIFF
--- a/src/components/collaboration-mode/collaboration-mode-picker.tsx
+++ b/src/components/collaboration-mode/collaboration-mode-picker.tsx
@@ -1,6 +1,7 @@
 import { Button, Popover, Icon } from '@wordpress/components';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
 import { paragraph } from '@wordpress/icons';
 
 import { viewIcon } from './view-icon';
@@ -23,14 +24,14 @@ interface ModeOption {
 const MODES: ModeOption[] = [
 	{
 		value: CollaborationMode.EDIT,
-		label: 'Edit',
-		description: 'Make changes',
+		label: __( 'Edit', 'vip-real-time-collaboration' ),
+		description: __( 'Make changes', 'vip-real-time-collaboration' ),
 		icon: paragraph,
 	},
 	{
 		value: CollaborationMode.VIEW,
-		label: 'View',
-		description: 'Focus on content',
+		label: __( 'View', 'vip-real-time-collaboration' ),
+		description: __( 'Focus on content', 'vip-real-time-collaboration' ),
 		icon: viewIcon,
 	},
 ];
@@ -69,6 +70,7 @@ export function CollaborationModePicker() {
 				icon={ currentIcon }
 				ref={ setPopoverAnchor }
 				iconSize={ 24 }
+				label={ __( 'Collaboration mode', 'vip-real-time-collaboration' ) }
 			/>
 			{ isPopoverVisible && (
 				<Popover
--- a/src/components/collaboration-mode/collaboration-mode-picker.tsx
+++ b/src/components/collaboration-mode/collaboration-mode-picker.tsx
@@ -1,0 +1,101 @@
+import { Button, Popover, Icon } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+import { paragraph } from '@wordpress/icons';
+
+import { viewIcon } from './view-icon';
+
+import '@/components/collaboration-mode/collaboration-mode-picker.scss';
+
+type CollaborationModeType = 'view' | 'edit';
+
+interface ModeOption {
+	value: CollaborationModeType;
+	label: string;
+	description: string;
+	icon: JSX.Element;
+}
+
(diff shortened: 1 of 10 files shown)


src/components/collaboration-mode/collaboration-mode-picker.tsx
```diff
@@ -0,0 +1,101 @@
+import { Button, Popover, Icon } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+import { paragraph } from '@wordpress/icons';
+
+import { viewIcon } from './view-icon';
+
+import '@/components/collaboration-mode/collaboration-mode-picker.scss';
+
+type CollaborationModeType = 'view' | 'edit';
```
ingeniumed: This'd be better in the settings store imo. It can be set to edit by default, and hidden outside of development mode.

We should ideally have an enum that's exported, and re-usable outside this component. We are going to be implementing the functionality and it'd be nice to prep for that too.

src/components/rtc-settings-panel.tsx
```diff
@@ -67,6 +68,9 @@ export function RTCSettingsPanel() {
 
 	return (
 		<>
+			<CollaborationMode>
```
ingeniumed: This should be feature flagged btw. It's not in use right now as we are still developing it.

ingeniumed (commented):
Should really use another slotfill that we have added to Gutenberg? I think there's some merit to the feedback that was mentioned in the Gutenberg PR. I know the designs suggest putting it in that location, but I wonder if it's better to instead have it in one of two places:

- Next to the publish/save button if it's really needed in the top document bar. There's some work in the workflow plugin that'd be useful to check out here on how to do this.

- Within the post info, where the details of the post are present. We could add a mode dropdown there so you could select what mode you are in.

I honestly like #2 more, as we would take advantage of the Post info and it relates to how the "post" is being worked on imo.

Image

Prob need some feedback from @jarekmorawski.

src/components/collaboration-mode/collaboration-mode-picker.tsx
ingeniumed: https://github.com/Automattic/vip-real-time-collaboration/compare/add/view-only-mode?expand=1

Got a version of this for the view only mode that I'm working on. It's a toggle for just the view only mode but ideally it should be a "enable elected modes" which in turn reveals that modal.

shekharnwagh:
I was thinking if toggling off the mode selection should reset the mode to the default value - `edit`, but maybe that would be confusing to the user as why was I moved to edit mode from view mode. This toggle to hide/show this feature might be temporary until we finish it, so my though is to not add additional logic for this.

ingeniumed:
> I was thinking if toggling off the mode selection should reset the mode to the default value - `edit`, but maybe that would be confusing to the user as why was I moved to edit mode from view mode. This toggle to hide/show this feature might be temporary until we finish it, so my though is to not add additional logic for this.

Yip, the toggle is temporary during the development of the elected modes feature. Once we finish development, we'll remove the toggle and allow the model to always be visible with a default value set to Edit. Then, as part of enforced modes we can tweak what the default model values should be.

src/store/collaboration-mode-store.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const reducer = (
+	state = DEFAULT_STATE,
+	action: CollaborationModeAction
+): CollaborationModeState => {
+	switch ( action.type ) {
+		case 'SET_MODE': {
+			return {
+				...state,
+				currentMode: action.payload.mode,
```
ingeniumed, line 30: Noting this for a follow up PR: We should save this mode into local storage like the settings are saved right now and read from there as well. That way on refresh it's not lost.

ingeniumed (approved):

src/store/collaboration-mode-store.ts
shekharnwagh: I didn't persist this in `localStorage` intentionally as that would lead to that getting persisted for other posts even those open in other tabs. Using `sessionStorage` might be better as thats limited to the current tab, would still apply to another post opened in same tab. If we do want to persist on refresh, might to strike a balance here. Top of mind we could persist something like `Record< postId, mode >` in `sessionStorage` and check by `postId`.

ingeniumed: I figured, whether we use local or session storage the key would be used to differentiate that. 

One other important part is that, we would want to eventually persist these on user meta possible so a user can carry this from one browser session to the other.

ingeniumed:
Noting that in https://github.com/Automattic/vip-real-time-collaboration/pull/110, there's a bunch of cleanup I'm doing to this PR.

PR #110, opened 2025-11-17
Add a view only mode - part 1

ingeniumed:
## Description

This'll be the first PR for the view only mode feature. 

It covers the following:

* Move the collaboration mode picker next to the collaborators list, and have it serve as the notification to the user for what mode they are in.
* Wire up the view only mode, to the actual view functionality.
* Implement partial view only functionality:
  * Make the code editor read only
  * Make the content of a post read only
  * Make the post title in a code editor read only
* Default to the current behaviour, where only the code editor is read only if it's not a dev environment

## ToDos for the follow up PR

* Make the title read only when the view only mode is implemented
* View awareness functionality in the code editor

That'll be coming in a follow up PR.

## Bonus

I discovered a bunch of react hooks that have been left behind, after all refactoring that we did of our various overlays. I've removed all of these unused hooks.

## Testing

- Pull down this PR, and our Gutenberg branch - wpvip/rtc-plugin
- Open up two browser session
- Enable collaboration modes
- Change one session to view only
- No changes should be possible to the body of the post
- Switch to the code editor and verify it's disabled
- Change this session to edit
- All of the restrictions should be gone, including to the code editor
- Disable collaboration modes
- Now only the code editor should be disabled

Note: Changes in the code editor will cause the post to disappear for the second browser session. That's a known problem right now, and hopefully I can address it in a follow up PR. 

## Screenshots/Screencasts

https://github.com/user-attachments/assets/9db71fb6-0b30-484c-af3e-a4ecf38011fb


ingeniumed:
@jarekmorawski - tagging you to review the UI change that I made compared to the figma mock ups.

CleanShot 2025-11-17 at 15 17 54

CleanShot 2025-11-17 at 15 17 48

ingeniumed:
Result of the design discussion: The dropdown will be moved back to the left, in the slot fill that @shekharnwagh added. That'll be the notice to the user what view they are in.

ingeniumed:
CleanShot 2025-11-18 at 10 54 24

ingeniumed:
This is good to review now @shekharnwagh.

I've left the locking of the post title in the visual editor, and the awareness features in the code editor for another PR.

src/hooks/use-modify-code-editor.ts
```diff
@@ -0,0 +1,69 @@
...
+		select => select( CollaborationModeStore ).getMode()
+	);
+
+	// The code editor is set to read-only in the following cases:
+	// 1. When the editor is in text mode.
+	// 2. When the Collaboration Mode Picker is enabled and the current mode is View.
+	// TODO: Remove the Collaboration mode enabled check once we complete the feature.
+	const shouldCodeEditorBeReadOnly =
+		editorMode === 'text' &&
+		( ! isCollaborationModeEnabled || collaborationMode === CollaborationMode.VIEW );
```
shekharnwagh, line 51: This makes it read-only when "Enable collaboration modes" toggle is off irrespective of the mode selected. From what I understand the default should be edit mode and only time it becomes read-only is when its explicitly set `View` and the "Enable collaboration modes" toggle is on.

shekharnwagh (approved):
Just had one comment, otherwise looks good. Thank you for working on this!

src/hooks/use-modify-code-editor.ts
ingeniumed: That's on purpose, because we haven't fully implemented this feature or released it. So if that toggle is off (which it'll be in production), this feature should be disabled. If it's disabled, that means it should work like it does in prod which is to always make the code editor read only. I've sprinkled in the TODOs to remove all this once we have fully released the feature.

PR #111, opened 2025-11-20
[VIP-2463] Add debug data export button to settings panel

shekharnwagh:
> Linear - [VIP-2463](https://linear.app/a8c/issue/VIP-2463)

Adds a new button in the RTC settings panel that exports the following debug data JSON with following structure -

```typescript
interface YDocDebugData {
	doc: {
		state: {
			version: number;
		};
		// YJS representation of Gutenberg entity
		document: {
			[key: string]: unknown;
		};
	};
	clients: {
		[clientId: string]: {
			id: {
				client: number;
				clock: number;
			};
			length: number;
			origin: string | null;
			rightOrigin: string | null;
			// YJS parent object
			parent: Record<string, unknown>;
			// key in the parent object for the current item
			parentSub: string;
			redone: Record<string, unknown> | null;
			content: Record<string, unknown>;
			info: number;
			// YJS left item reference
			left: Record<string, unknown> | null;
			// YJS right item reference
			right: Record<string, unknown> | null;
		}[];
	};
	// Map of the YJS client ID to the WordPress user as YJS client ID is re-generated on each refresh
	userMap: {
		[clientId: string]: {
			wpUserId: number;
			name: string;
			email: string;
		};
	};
}
```

This would be useful in understanding how the Post blocks/attributes are updated in a collaborative session by participants.

### Screenshots

Screenshot 2025-11-25 at 10 38 29 PM

### DIFF
--- a/src/awareness-manager.ts
+++ b/src/awareness-manager.ts
@@ -134,9 +134,8 @@ export class AwarenessManager {
 	}
 
 	public static getDebugData(): YDocDebugData | null {
-		const logger = new Logger( 'awareness-manager' );
 		if ( ! AwarenessManager.__instance?.awareness?.doc ) {
-			logger.error( 'getDebugData() awareness document not found' );
+			AwarenessManager.__instance.logger.error( 'getDebugData() awareness document not found' );
 			return null;
 		}
 
--- a/src/awareness-manager.ts
+++ b/src/awareness-manager.ts
@@ -47,9 +47,36 @@ interface AwarenessStateChange {
 	removed: AwarenessClientID[];
 }
 
+// Type for serializable left/right item references to avoid deep nesting
+type SerializableYItemRef = Pick< Y.Item, 'id' | 'length' | 'origin' | 'content' >;
+
+// Serializable Y.Item - only includes data properties with shallow left/right references
+type SerializableYItem = Pick<
+	Y.Item,
+	| 'id'
+	| 'length'
+	| 'origin'
+	| 'rightOrigin'
+	| 'parent'
+	| 'parentSub'
+	| 'redone'
+	| 'content'
+	| 'info'
+> & {
+	left: SerializableYItemRef | null;
+	right: SerializableYItemRef | null;
+};
+
 interface YDocDebugData {
 	doc: Record< string, unknown >;
-	clients: Map< number, unknown >;
+	clients: Record< number, Array< SerializableYItem > >;
+}
+
+/**
+ * Type guard to check if a struct is a Y.Item (not Y.GC)
+ */
+function isYItem( struct: Y.Item | Y.GC ): struct is Y.Item {
+	return 'content' in struct;
 }
 
 export class AwarenessManager {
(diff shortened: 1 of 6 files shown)


src/components/debug-data-export-button.tsx
```diff
@@ -0,0 +1,38 @@
+import { downloadBlob } from '@wordpress/blob';
+import { Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import safeStringify from 'safe-stringify';
+
+import { AwarenessManager } from '@/awareness-manager';
+
+export function DebugDataExportButton() {
+	const handleDownload = () => {
```
chriszarate: this function can be in outer scope

src/awareness-manager.ts
```diff
@@ -93,6 +125,50 @@ export class AwarenessManager {
 		);
 	}
 
+	public static getDebugData(): YDocDebugData | null {
+		const logger = new Logger( 'awareness-manager' );
```
chriszarate: Can we create this once in outer scope and remove `this.logger`?

src/awareness-manager.ts
```diff
@@ -93,6 +125,50 @@ export class AwarenessManager {
...
+			return null;
+		}
+
+		const ydoc = AwarenessManager.__instance.awareness.doc;
+
+		// Manually extract doc data to avoid deprecated toJSON method
+		const docData: Record< string, unknown > = {};
+		ydoc.share.forEach( ( value, key ) => {
+			docData[ key ] = value.toJSON();
+		} );
```
chriszarate:
```suggestion
		const docData: Record< string, unknown > = Object.fromEntries(
			Object.entries( ydoc.share ).forEach( ( [ value, key ] ) => [ key, value.toJSON() ] )
		);
```

src/awareness-manager.ts
```diff
@@ -111,6 +187,39 @@ export class AwarenessManager {
...
+		// Persist Yjs client ID to WordPress user mapping in the YDoc
+		// This ensures the mapping is synced and available even after users disconnect
+		this.persistUserClientsMap( this.awareness.clientID, userInfo );
+	}
+
+	/**
+	 * Persist a user client mapping in the YDoc.
+	 * Stores Yjs client ID -> WordPress user info mapping for debugging.
+	 */
+	private persistUserClientsMap( yjsClientId: number, userInfo: UserInfo ): void {
```
chriszarate: Instead of storing these on the Yjs state map, I think each client can just keep an in memory map of all the clients they have seen. Not perfect (misses users that the current client has not seen), but perhaps good enough for now

https://github.com/Automattic/vip-real-time-collaboration/commit/e3c6eb2698a6c4af5f905a40845bd6a98902663c

chriszarate (approved):
I pushed a couple of cleanup commits. Looks great!

src/components/rtc-settings-panel.tsx
```diff
@@ -143,10 +144,6 @@ export function RTCSettingsPanel() {
 						{ __( 'Notifications', 'vip-real-time-collaboration' ) }
 					</Heading>
 
-					<Heading level={ 2 } style={ { marginBottom: '2px' } }>
```
ingeniumed, line 146: Why was this removed?

chriszarate: I thought it was a mistake. It's a level 2 heading following a level 3 and looked anomalous. It's also the only sub-section so provided no distinction.

ingeniumed: It was done on purpose, to allow for the notifications to be categorized based on their purpose. I think the User sub-section was removed? 

There were supposed to be 2 of em:

- Post: This would be for save/publish and any other post specific action
- Collaborator: This would be for when a collaborator enters/exits a post

I think because the second one was removed it looks confusing. Lemme dig that up, and re-add it because the toggles don't make a lot of sense right now. For example, "enters" doesn't quite convey what that means.

ingeniumed: It was removed in https://github.com/Automattic/vip-real-time-collaboration/commit/7eca8cda759b7cf6d8008bf67e4687215ea60475.

I'll add it back in.

shekharnwagh: > It was removed in https://github.com/Automattic/vip-real-time-collaboration/commit/7eca8cda759b7cf6d8008bf67e4687215ea60475.

My bad, I confused it with the new collaborators section which we added in that PR.

chriszarate: Can you fix the heading levels?

ingeniumed: Noting that this has been fixed in #120

PR #113, opened 2025-11-21
Update view mode

ingeniumed:
### Description

I've moved the collaborator mode to GB, within the core-data store. The dropdown and it's related functionality will now be powered by that new property.

Also, I've set the collaborator mode dropdown to be on by default when in development otherwise its off. That way, we don't need a separate toggle just for turning this on and off locally.

There are 3 view functionalities offered by our plugin:

- The code editor is read only in view mode, and in production. It's enabled in edit mode.
- Editing any blocks is disabled, including interacting with any properties of it.
- Disable interaction with anything in the sidebar

### What I haven't been able to do

Try to allow blocks to be interacted with, but not change any properties of it while also syncing the UIs. Right now, the UI doesn't sync which creates a very bad experience.

### Screencast

https://github.com/user-attachments/assets/8ceda668-392b-48b9-85a5-f22ca2d25537


src/hooks/use-disable-sidebar-interaction.ts
```diff
@@ -0,0 +1,48 @@
...
+		select( rtcSettingsStore ).getSetting( Setting.COLLABORATION_MODE_PICKER )
+	);
+
+	const shouldSidebarBeReadOnly =
+		isCollaborationModeEnabled && currentCollaborationEditorMode === 'view';
+
+	// Manage sidebar interaction.
+	useEffect( () => {
+		const sidebarElement = document.querySelector( '.editor-sidebar' );
+		const headerSettingsElement = document.querySelector( '.editor-header__settings' );
```
ingeniumed: I did try locking post saving, and auto saving instead of doing this. That is not the best solution because if the pre-publish panel is enabled, then any changes made there result in UI quirks. This could include a tag not saving for example which wouldn't be the best experience.

The thought I had is that, we can slowly remove each CSS tweak as we make changes within Gutenberg to allow native "read only" modes everywhere.

src/store/settings-store.ts
```diff
@@ -36,7 +36,7 @@ const DEFAULT_STATE: SettingsState = {
 	[ Setting.POST_UPDATE_NOTIFICATION ]: true,
 	[ Setting.USER_ENTER_NOTIFICATION ]: true,
 	[ Setting.USER_EXIT_NOTIFICATION ]: false,
-	[ Setting.COLLABORATION_MODE_PICKER ]: false,
+	[ Setting.COLLABORATION_MODE_PICKER ]: isDevelopment() ? true : false,
```
ingeniumed: This is not accounting for what happens when the value is stored in settings and re-used.

Ideally, it'd be better to just have this be hardcoded in code instead.

src/store/settings-store.ts
```diff
@@ -55,6 +58,12 @@ const reducer = (
 ): SettingsState => {
 	switch ( action.type ) {
 		case 'SET_SETTING': {
+			// ToDo: Delete COLLABORATION_MODE_PICKER once we complete the feature.
+			// Skip saving COLLABORATION_MODE_PICKER as it's computed dynamically
+			if ( action.payload.setting === Setting.COLLABORATION_MODE_PICKER ) {
```
ingeniumed, line 63: @chriszarate - This should rectify the bug that you were seeing. 

I'm dynamically computing the value for COLLABORATION_MODE_PICKER, wherein it's true in development and false otherwise. Considering we are developing this feature locally, it should always be on so it's easy to work on. We would get rid of this setting, and it's code once the feature is live.

IMO, we should do the same for debug tools as well so it's always on in development.

ingeniumed: If you have an alternate approach that you prefer lemme know. I didn't want to overtly complicate this, because this code will be deleted soon.

PR #119, opened 2025-12-03
Specify algorithm in WebSocket JWT verification

shekharnwagh:
## Summary

Improves WebSocket JWT authentication security by explicitly specifying the allowed algorithm during token verification.

## Why

Without explicitly specifying the algorithm, the JWT library accepts whatever algorithm is declared in the token header. This allows attackers to manipulate the algorithm field to bypass signature verification.

By explicitly requiring HS256, we ensure only tokens signed with our expected algorithm are accepted.

### DIFF
--- a/websocket-server/auth.ts
+++ b/websocket-server/auth.ts
@@ -36,7 +36,7 @@ function verifyToken( token: string | null | undefined, secret: string ): SyncTo
 		throw new Error( 'Missing token' );
 	}
 
-	const jwtPayload = jwt.verify( token, secret );
+	const jwtPayload = jwt.verify( token, secret, { algorithms: [ 'HS256' ] } );
 	if ( ! isSyncTokenPayload( jwtPayload ) ) {
 		throw new Error( 'Invalid JWT payload' );
 	}
--- a/websocket-server/auth.ts
+++ b/websocket-server/auth.ts
@@ -1,6 +1,6 @@
-import jwt, { type JwtPayload } from 'jsonwebtoken';
+import { type JwtPayload } from 'jsonwebtoken';
 
-import { getRequestPathname } from './utils';
+import { getRequestPathname, verifyJwtToken } from './utils';
 
 import type { IncomingMessage } from 'node:http';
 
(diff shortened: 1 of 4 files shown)


chriszarate (changes requested):
Given this is a security vulnerability that we don't want to repeat, we might want to approach token validation with even more caution. Perhaps we can define wrapper functions for token operations in `utils.ts` that prevent this issue via extra type safety. Something like:

```ts
import jwt, { type Jwt, type JwtPayload, type VerifyOptions } from 'jsonwebtoken';

type WithRequired< T, K extends keyof T > = Omit< T, K > & Required< Pick< T, K > >;

type SignOptionsWithAlgo = WithRequired< VerifyOptions, 'algorithms' >;

export function verifyJwtToken(
	token: string,
	secret: string,
	options: SignOptionsWithAlgo
): JwtPayload | Jwt | string {
	return jwt.verify( token, secret, options );
}
```

Additionally, we could leverage eslint's [`no-restricted-imports` rule](https://eslint.org/docs/latest/rules/no-restricted-imports) to ban importing `jwt` with a pointer to the `utils` file. Then override the rule in that one spot.

shekharnwagh:
> Given this is a security vulnerability that we don't want to repeat, we might want to approach token validation with even more caution. Perhaps we can define wrapper functions for token operations in `utils.ts` that prevent this issue via extra type safety. 

Pushed a commit e63e456 to address this. Thank you for the suggestion !

websocket-server/auth.ts
```diff
@@ -36,7 +36,7 @@ function verifyToken( token: string | null | undefined, secret: string ): SyncTo
 		throw new Error( 'Missing token' );
 	}
 
-	const jwtPayload = jwt.verify( token, secret );
+	const jwtPayload = verifyJwtToken( token, secret, { algorithms: [ 'HS256' ] } ) as unknown;
```
chriszarate, line 39: cast is unnecessary 

```suggestion
	const jwtPayload = verifyJwtToken( token, secret, { algorithms: [ 'HS256' ] } );
```

chriszarate (approved):
nice, thanks!

PR #122, opened 2025-12-05
Add the plugin version information to the plugin settings

ingeniumed:
### Description

The problem we face right now is that, it's difficult to tell if the right version of Gutenberg is being used alongside VIP RTC or not. #68 was an attempt at solving this, but that wasn't the right approach given we want all our changes to go into core.

Instead, we are now going to show all the information in the plugin settings page. That way, customers and other users can easily tell what versions of the plugins they are using. We can then use for debugging or reporting issues.

### Development mode

CleanShot 2025-12-05 at 09 47 46

### Production mode

CleanShot 2025-12-05 at 09 49 33

### DIFF
--- a/inc/Settings/Settings.php
+++ b/inc/Settings/Settings.php
@@ -62,7 +62,7 @@ final class Settings {
 		add_settings_section(
 			'plugin-settings',
 			'',
-			[ __CLASS__, 'display_settings_instructions' ],
+			'__return_null',
 			self::SETTINGS_PAGE_SLUG
 		);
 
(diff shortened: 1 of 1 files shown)


ingeniumed:
I kept things simple as we don't have a lot on the plugin settings page at the moment. I kept the design similar to the governance plugin, and focused on just showing what we think is important. 

Happy to change this as needed.

inc/Settings/Settings.php
```diff
@@ -154,11 +176,36 @@ public static function display_settings_radio( array $args ): void {
...
+		$value = constant( $constant_name );
+
+		return $value;
+	}
+
+	public static function get_vip_rtc_version(): string {
+		return self::get_string_constant( 'VIP_REAL_TIME_COLLABORATION__PLUGIN_VERSION' );
+	}
+
+	public static function get_gutenberg_version(): string {
```
ingeniumed, line 199: Both this, and the commit sha are only available when Gutenberg is built for production. This isn't the case when `npm run dev` is done which is why I've accounted for that.

inc/Settings/Settings.php
```diff
@@ -101,14 +101,36 @@ public static function add_options_page(): void {
...
+			<h2><?php esc_html_e( 'Plugin Settings', 'vip-real-time-collaboration' ); ?></h2>
 			<form action="options.php" method="post">
 				<?php
 				settings_fields( self::SETTINGS_PAGE_SLUG );
 				do_settings_sections( self::SETTINGS_PAGE_SLUG );
 				submit_button();
 				?>
 			</form>
+
+			<h2><?php esc_html_e( 'Plugin Debug Information', 'vip-real-time-collaboration' ); ?></h2>
```
chriszarate: Since not all of this pertains to the plugin
```suggestion
			<h2><?php esc_html_e( 'Debug Information', 'vip-real-time-collaboration' ); ?></h2>
```

chriszarate (approved):

PR #124, opened 2025-12-05
Handle errors in getUserEmail with graceful fallback

shekharnwagh:
Wrap user email fetching in a try-catch to prevent unhandled promise rejections if the WordPress core-data store fails to retrieve user info. Log any errors and return an empty string to ensure undefined does not propagate to the UI. If the email cannot be fetched, collaborators are displayed using only display names.


chriszarate:
As discussed out-of-band, we will separately remove email from user info.

PR #127, opened 2025-12-10
Add type safety and flexibility to awareness

chriszarate:
This is a major refactor of our approach to awareness, with the following goals:

1. Improve type safety of awareness in general. Yjs awareness states are schema-less JSON blobs and the code provides no type bindings. Introducing `TypedAwareness` allows us to set and get user states with enforced types.
1. Address the mistake of mixing synced shared state (`UserInfo`) with private, local state (`isMe`, `isDisconnected`). This was another source of confusion and bugs. The latter should be "mixed in" only at the last minute when the state is ready to be used by our components.
1. Remove the complexity of a reactive (Redux) store on top of an event-driven awareness implementation. This was the source of much confusion and many bugs. Instead of a Redux store, we can provide a React hook wrapping `useState` (`usePostEditorAwarenessState`) and provide a simple subscription binding to update that state when needed.
1. Address the flawed, hard-coded assumption that the first-loaded post entity represents the post in the post editor. As we expand RTC to other places in WordPress admin, this assumption would fail and hold us back. Instead, we should support multiple awareness instances and let the editor components tell us which awareness state they are interested in.
1. Create our own awareness instances instead of expecting them to be created by Gutenberg.

`PostEditorAwarenessState` is the first implementation of this newer approach to awareness, with others expected in the future—e.g., `PostListAwarenessState`, `SettingsPageAwarenessState`, etc. I recommend that you start with the classes defined in `awareness-state.ts` and `awareness-types.ts` and follow their public methods as they are used externally.

Apologies for the size of this pull request. Know that much of this code has not changed and only relocated and enhanced with type safety. There are relatively few changes to non-awareness code.

Note for @alecgeatches: The changes to `use-render-cursors.ts` expose the fact that we need access to the Yjs doc deep within the component tree, which feels wrong. In the future, perhaps we can move more of this logic to `PostEditorAwarenessState` or awareness utility functions.


src/awareness/awareness-state.ts
```diff
@@ -0,0 +1,295 @@
...
+				} )
+		);
+
+		if ( ! forceUpdate ) {
+			if ( areMapsEqual( this.previousSnapshot, updatedStates, this.isStateEqual.bind( this ) ) ) {
+				this.logger.debug( 'Awareness state unchanged, not updating subscribers.' );
+				return;
+			}
+
+			this.logger.debug( 'Awareness state changed, updating subscribers.' );
```
ingeniumed: Noting the same here as well.

src/awareness/awareness-state.ts
```diff
@@ -0,0 +1,295 @@
...
+		} else {
+			this.logger.debug( 'Forcing awareness store update.' );
```
ingeniumed: This one happens when a new peer joins, with remaining peers. Same thing here - is this one needed?

Feel like yes for this.

src/awareness/awareness-state.ts
```diff
@@ -0,0 +1,295 @@
...
+						isMe,
+					};
+
+					return [ clientId, state ];
+				} )
+		);
+
+		if ( ! forceUpdate ) {
+			if ( areMapsEqual( this.previousSnapshot, updatedStates, this.isStateEqual.bind( this ) ) ) {
+				this.logger.debug( 'Awareness state unchanged, not updating subscribers.' );
```
ingeniumed: Noting that this currently is filling up the console logs.

Is this needed? Feel like no given how much is in the logs.

ingeniumed (approved):
The PR looks good, only thing to note is the excessive console logging that I'm now seeing. I've noted them below.

src/awareness/awareness-state.ts
chriszarate: Good catch, meant to remove these after my testing

src/awareness/awareness-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+			{
+				email: userState.userInfo.email,
+				name: userState.userInfo.name,
+				wpUserId: userState.userInfo.id,
+			},
+		] )
+	);
+
+	// Serialize Yjs client items to avoid deep nesting
+	const serializableClientItems: Record< number, Array< SerializableYItem > > = {};
```
alecgeatches, line 64: I'm curious what `serializableClientItems` holds at the end of this process, especially with a little magic like `Y.GC` garbage collection items being removed. Can you maybe add a bit longer of a comment here about what the result is?

src/awareness/post-editor-awareness-state.ts
```diff
@@ -0,0 +1,158 @@
...
+	type SelectableBlock,
+	updateSelectionInEntityRecord,
+} from '@/utilities/selection';
+import { areEditorStatesEqual, areUserInfosEqual } from '@/utilities/user';
+
+import type { PostEditorState, UserInfo } from '@/awareness/awareness-types';
+import type { WPBlockSelection } from '@wordpress/editor/build-types/store/selectors';
+
+export class PostEditorAwarenessState extends AwarenessState< PostEditorState > {
+	protected equalityFieldChecks = {
```
alecgeatches, line 34: I love this type-enforced system for checking field equality. Great thinking!

src/awareness/post-editor-awareness-state.ts
```diff
@@ -0,0 +1,158 @@
...
+	private updateSelectionState(
+		selectionStart: WPBlockSelection,
+		selectionEnd: WPBlockSelection
+	): void {
+		const ydoc = this.doc.getMap( RECORD_KEY );
+		const yBlocks = ydoc.get( 'blocks' ) as Y.Array< SelectableBlock >;
+		const selection = getSelectionState( selectionStart, selectionEnd, yBlocks );
+
+		// Debounce remote awareness updates.
+		this.setLocalStateFieldWithDebounce(
```
alecgeatches: @chriszarate I'm seeing different results for rapid selection changes between `trunk` and this branch. These may have been on purpose. On `trunk`, if we send a lot of selection updates, they come through on other browsers:

https://github.com/user-attachments/assets/223ec640-519f-49c1-ab24-8fc55b32fc5a

However in this PR, if I try the same test, I only get the first selection update and then none until selection stops changing:

https://github.com/user-attachments/assets/49f1891d-35f6-454f-82e1-de2a77419c9f

I think this is due to a debounce over a throttle approach. I'd prefer the first behavior, since we will at least send updates for in-between states regularly.

alecgeatches (approved):
Fantastic work, the AwarenessManager has been a source for several issues and hidden inefficiencies. The typing you've added to properties and equality functions are a great way to work around those, plus less complexity in managing a store.

In testing locally everything works well, with one minor throttle/debounce difference I noted above. Thank you!

src/awareness/awareness-utils.ts
chriszarate: This is unmodified relocated code, but good suggestion. I'll take a look

src/awareness/post-editor-awareness-state.ts
chriszarate: The code previously took a debounce approach, but it might not have been working 100%. I'll take a look

src/awareness/awareness-utils.ts
chriszarate: I don't think I can speak authoritatively on this. I will kindly ask @shekharnwagh to take this on in a separate PR.

src/awareness/post-editor-awareness-state.ts
chriszarate: Changed to throttling in 95f19ab

src/awareness/awareness-state.ts
chriszarate: Removed in [5436d73](https://github.com/Automattic/vip-real-time-collaboration/pull/127/commits/5436d73a081fc0594dd960b87dc0af596bb584fb)

PR #129, opened 2025-12-11
Update the collaborate mode's store

ingeniumed:
### Description

This is a companion PR to [this PR](https://github.com/WordPress/gutenberg/pull/73898). 

It'll move the collaborate mode calls to come from the core-data store instead of the editor store.

An important change: The collaboration mode picker is locked down when the enforced mode is set to prevent anyone from changing the mode they are in.

### Screencast

https://github.com/user-attachments/assets/cc1c464f-827e-4e30-a6ed-77440a173cd4

### DIFF
--- a/src/components/collaboration-mode-picker.tsx
+++ b/src/components/collaboration-mode-picker.tsx
@@ -56,27 +56,27 @@ export function CollaborationModePicker() {
 	const [ popoverAnchor, setPopoverAnchor ] = useState< HTMLElement | null >( null );
 
 	const { selectedCollaborationEditorMode, shouldPickerBeLockedDown } = useSelect( select => {
-		const { getCurrentPostType, getCurrentPostId } = select( editorStore ) as EditorStoreSelectors;
-		const { getEditedEntityRecord, getCollaboratorMode } = select( coreStore ) as CoreDataSelectors;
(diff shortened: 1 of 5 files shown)


src/components/collaboration-mode-picker.tsx
```diff
@@ -41,11 +55,33 @@ export function CollaborationModePicker() {
...
+
+		if ( postType && postId ) {
+			// This mirrors the way it's done in the switchEditorMode action.
+			const entityRecord = coreSelectors.getEditedEntityRecord( 'postType', postType, postId ) as
+				| EntityRecordWithRtcMeta
+				| undefined;
+			const rtcMeta = entityRecord?.meta?.rtc;
+			if ( rtcMeta?.enforcedModeOwner && rtcMeta?.enforcedMode === 'codeEditor' ) {
+				isLockedDown = true;
+			}
```
chriszarate, line 75: Why isn't this purely a function of collaborator mode state from the store? Having sync logic like this in editor components is not ideal. Do we need another property like `isEnforced` in the store?

ingeniumed:
Since we are punting the collaborator modes (enforced/elected modes) feature to post 7.0, I'm closing this PR.

We can revisit this post 7.0.

PR #132, opened 2025-12-16
Improve React linting and simplify render cursor logic

chriszarate:
React code was not being fully linted because React was not installed as a (dev) dependency.


alecgeatches (approved):
Tested locally and cursors seem to be rendering as expected in all of the device sizes and "Show Template" mode. Looks good to me!

PR #133, opened 2025-12-16
Separate `getGetAbsolutePositionIndex`

chriszarate:
Separate `getGetAbsolutePositionIndex` from post editor state object. It does not have a dependency on the current awareness state, just the current awareness instance.

A separate, memoized accessor leads to fewer rerenders.


chriszarate:
Code has moved to Gutenberg `trunk`.

PR #134, opened 2025-12-17
Fix editor crash when opening Yjs inspector

alecgeatches:
## Reproduction

1. Open a post in a local environment with the Yjs inspector enabled.
2. Click the Yjs inspector link and open it in a new tab.
3. Wait a few seconds.
4. See the editor crash:

https://github.com/user-attachments/assets/d3611a98-4ee6-4c63-b01e-676a68a6b89e

```
Uncaught TypeError: can't access property "avatarUrl", userInfo is undefined
    Avatar avatar.tsx:28
```

## Cause

When the Yjs inspector opens, it also connects to the awareness instance on the post entity. Our code currently expects awareness users to provide a minimum set of user data in `userInfo` (like `avatarUrl`) to be used in rendering that the inspector does not provide.

## Fix

Filter out invalid user states without a `userInfo` property.

### DIFF
--- a/src/awareness/awareness-state.ts
+++ b/src/awareness/awareness-state.ts
@@ -253,7 +253,6 @@ export abstract class AwarenessState<
 
 		const updatedStates = new Map< number, EnhancedState< State > >(
 			[ ...this.disconnectedUsers, ...states.keys() ]
-				.filter( clientId => this.seenStates.has( clientId ) )
 				.filter( clientId => {
 					// Exclude any users without `userInfo`.
 					// This can happen from the Yjs inspector, which joins the awareness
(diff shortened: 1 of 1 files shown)


src/awareness/awareness-state.ts
```diff
@@ -254,6 +254,12 @@ export abstract class AwarenessState<
 		const updatedStates = new Map< number, EnhancedState< State > >(
 			[ ...this.disconnectedUsers, ...states.keys() ]
 				.filter( clientId => this.seenStates.has( clientId ) )
```
chriszarate: covered by following filter
```suggestion
```

chriszarate (approved):
thanks!

PR #135, opened 2025-12-22
Add soft limit to prevent partial client connections

shekharnwagh:
Clients need multiple WebSocket connections for different entities. When the server reaches max capacity, new connections are rejected, causing some entities to connect while others fail, resulting in a broken experience.

This PR introduces a soft connection limit that reserves buffer connections for existing clients to complete their setup. When total connections reach the soft limit, only existing clients can connect (new clients are rejected), ensuring atomic behavior - a client connects all their entities or none. The soft limit is calculated as `max - buffer` and is configurable via the `VIP_RTC_WS_BUFFER_CONNECTIONS` environment variable. All state is derived from `wss.clients` to prevent drift from edge cases where close handlers don't fire.

For semantic clarity, `connectionId` has been renamed to `wpClientId` throughout the codebase, since the identifier represents a client (browser tab) rather than individual connections. The API maintains backwards compatibility by accepting both `wpClientId` (new) and `connectionId` (deprecated) parameters during the transition period.


websocket-server/index.ts
```diff
@@ -72,13 +82,14 @@ setPersistence( new NoopPersistenceProvider() );
 wss.on( 'connection', ( ws: WebSocket, request: IncomingMessage ) => {
 	const connectionStartTime = Date.now();
 	const connectionId = getConnectionId( request, JWT_SECRET );
```
chriszarate: So the `connectionId` is really more of a `clientId`—a unique ID computed once (in-memory) by the client. Perhaps we should rename it for clarity?

websocket-server/client-connections.ts
```diff
@@ -0,0 +1,143 @@
...
+		this.soft = this.calculateSoftLimit( limits.max );
+	}
+
+	/**
+	 * Calculate soft limit from max connections
+	 * - If max < 50: soft = max (disabled)
+	 * - Otherwise: soft = 90% of max
+	 * - Buffer (max - soft) never exceeds 50
+	 * This ensures that a large number of connections are not tied up as buffer and thus reducing
+	 * the number of connections that can be established.
```
chriszarate: I'm not sure the percentage approach is that helpful. It seems like we want a soft limit that reserves enough space for one client's connection needs. If that number is 50, then we should try to reserve that amount. If reserving 50 is not practical, then we should disable and warn, because the soft limit is not working as intended.

So the logic would be:

```ts
if ( max < buffer *2 ) {
   logger.warn( 'Soft limit disabled...' );
   return max;
}

return max - buffer;
```

What do you think?

chriszarate: And the buffer value should probably be configurable via env var

websocket-server/client-connections.ts
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	private calculateSoftLimit( max: number ): number {
+		// Disable soft limit for small connections limits
+		if ( max < 50 ) {
+			return max;
+		}
+
+		const SOFT_LIMIT_PERCENTAGE = 0.9;
+		const MAX_BUFFER = 50;
```
chriszarate: This "buffer" represents the number of connections we expect a client to open. What is a better name?

websocket-server/index.ts
```diff
@@ -72,13 +82,14 @@ setPersistence( new NoopPersistenceProvider() );
...
+	clientConnectionStore.addConnection( connectionId );
 
 	/**
 	 * Set up the connection
 	 */
 	setupWSConnection( ws, request );
 
-	recordConnectionOpen( connectionId );
+	recordConnectionOpen( connectionId, clientConnectionStore.getActiveClientCount() );
```
chriszarate: why not continue to rely on `wss.clients.size`, which should always be accurate? has it not yet incremented for the current connection?

shekharnwagh: That would be used for total connections, while `clientConnectionStore.getActiveClientCount()` would identify number of clients - connections grouped by connectionId/clientId

chriszarate: Nevermind, I understand the distinction now. Confusing terms!

I might prefer keeping the accounting of these separate. We already have an interval that keeps the active connections gauge in sync using `wss.clients.size`. Why introduce another bookkeeping method?

Edit for clarity: Can we introduce another interval that keeps active clients gauge updated?

websocket-server/index.ts
shekharnwagh: Yes, makes sense. IIRC, I avoided using `clientId` due it being used by both YJS and Gutenberg blocks and ended up using `connectionId` then to avoid confusion. Now, it makes more of a case for using it, so will change it.

websocket-server/client-connections.ts
shekharnwagh: Makes sense and we haven't yet encountered any issues yet due to this. It provides a good starting point that we can refine further.

websocket-server/client-connections.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * Client connection store for managing active connections per client
+ */
+export class ClientConnectionStore {
+	/**
+	 * Map of connection_id to count of active connections for that client
+	 */
+	private clientConnections = new Map< string, number >();
```
chriszarate: The map value (connections per client) is unused. Why not `boolean`?

shekharnwagh: We do use it here to delete the entry when the last connection for the client closes - https://github.com/Automattic/vip-real-time-collaboration/pull/135/changes#diff-99ad138e42f42b7efbba03a53bf4b95711e9e16cb3868857dad9be9e8f9cab7aR87-R98

websocket-server/index.ts
chriszarate: Maybe a third term is needed then, like `wpClientId`?

websocket-server/index.ts
shekharnwagh: > Can we introduce another interval that keeps active clients gauge updated?

I am not sure what benefit another interval would do. The gauge for all connections - `activeConnectionsGauge` has an independent source `wss.clients.size` which can be used to reconcile it. The gauge `activeClientsGauge` for tracking clients (connections de-duplicated by connectionId/clientId) doesn't have another source. We keep updating every time connection opens and closes, so I don't see a need for an interval. OR do you mean we should update it bases on an interval instead of those open and close connection events ?

chriszarate: The latter

websocket-server/client-connections.ts
shekharnwagh: Usually, only a single client would be affected if it can't secure all necessary connections, but in rare cases - like multiple clients opening posts at once, more users could be impacted. To ensure a stable experience, I thought it safer to use a larger buffer rather than a smaller one (like 5), since the difference is negligible when supporting high limits (e.g., 1000+ connections). This edge case may be somewhat more common on large multisites. That said, if we don't want to worry about these edge cases, we could simply set the buffer to something like 2–4 times the typical connections per client, so it doesn't need frequent adjustment as we add sync support for more entities.

shekharnwagh: > What is a better name?
Maybe something like `reserveConnections` ?

websocket-server/index.ts
shekharnwagh: The current approach leverages `recordConnectionOpen`, which already updates the gauge for total connections in real-time. Even if we switched to using an interval to track active clients, we’d still need to maintain the state within the `ClientConnectionStore`. Essentially, the choice is between recording active client counts immediately as connections open or close, versus updating this metric periodically via an interval. Is there a specific benefit to using the interval-based method instead of real-time updates?

websocket-server/client-connections.ts
chriszarate: Buffer is fine per your explanation. Maybe add some of that thinking around in a comment above the const definition?

websocket-server/index.ts
chriszarate: > Essentially, the choice is between recording active client counts immediately as connections open or close, versus updating this metric periodically via an interval. Is there a specific benefit to using the interval-based method instead of real-time updates?

Can you think of any scenarios where the `recordConnectionClose` function does not fire, and thus the active client count becomes inaccurate over time—because clients never reach `0` connection count? That is my main worry: that bugs or edge cases in the management of `clientConnections` make this bookkeeping method less accurate.

If you feel confident, then proceed as written.

If you have concerns, and we expect `wss.clients` to always be accurate, structuring our bookkeeping around that might be better. We can extend the `WebSocket` type and store `.wpClientid` on it when the connection opens, then get the active client count with

`new Set( Array.from( wss.clients ).map( ws => ws.wpClientId ) ).size`

or you could use a `WeakMap` or something. Just my thoughts, happy to go with whatever you think here.

inc/Api/AuthApiController.php
```diff
@@ -77,19 +77,19 @@ public function get_auth_token( WP_REST_Request $request ): WP_REST_Response|WP_
 		$sync_object_type = $request->get_param( 'syncObjectType' );
 		$sync_object_id = $request->get_param( 'syncObjectId' );
 		/** @psalm-suppress MixedAssignment */
-		$connection_id = $request->get_param( 'connectionId' );
+		$wp_client_id = $request->get_param( 'wpClientId' );
```
chriszarate, line 89: I think we want some fallback code here so that we accept either during the rollout period.

chriszarate (approved):

inc/Api/AuthApiController.php
shekharnwagh: Pushed a change for this - 71c796b

shekharnwagh:
@chriszarate Refactored the new code to functions instead of a class to keep it simple since we aren't managing any state there - 554808d. Doesn't change the functionality but wanted to loop you in

PR #137, opened 2026-01-06
Remove the collaborator mode feature

ingeniumed:
### Description

Since we are punting the collaborator modes feature, this PR is necessary to reverse the experimental code that we added in for this feature.

This is a companion PR to https://github.com/WordPress/gutenberg/pull/74374.


src/hooks/use-modify-code-editor.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+import { useSelect } from '@wordpress/data';
+import { store as editPostStore, type EditPostStoreSelectors } from '@wordpress/edit-post';
+import { useEffect } from '@wordpress/element';
+
+/**
+ * Custom hook that manages read-only state for the code editor.
+ * Makes the code editor textarea read-only by default.
+ */
+export function useModifyCodeEditor() {
```
ingeniumed, line 12: I kept the read-only code editor feature as a hook, and the addition of the title being locked down as well. Before the collaborator modes feature, it was a component and the title wasn't targeted. This is a lot better.

PR #138, opened 2026-01-08
Adding filter for post type

smithjw1:
This PR adds a filter that lets customers exclude specific post types from real-time collaboration.


chriszarate:
The `get_supported_post_types()` method is no longer used and this new filter will have no effect. Discussion on upstream PRs led to a decision to enable collaboration by default without customization.

Filtering user permissions are a better strategy at this time.

smithjw1:
I've opened #139 as an alternative.

PR #142, opened 2026-01-19
Removing self-awareness

smithjw1:
We've talked about removing this setting for a while, but we just haven't gotten around to it.


chriszarate (approved):
very good

maxschmeling (approved):

PR #143, opened 2026-01-21
Update createWebSocketConnection function signature

alecgeatches:
Fixes a plugin error `provider.on is not a function` caused by error handling type changes in https://github.com/WordPress/gutenberg/pull/74075. The root cause is that the `ProviderCreator` type changed from an argument list to [an options object](https://github.com/Automattic/gutenberg/blob/70092462d4509b8d41c59b31fcf2fac0c9e0e504/packages/sync/src/types.ts#L110-L115).

Additionally, this PR starts adding support for the Gutenberg-based error dialog by:

1. Adding new types from the sync package
2. Returning an `on()` function that the websocket provider sends events to

This fixes the immediate problem loading posts, but there are still some unimplemented parts:

1. Removal of the local `<PostLockedModal>`, which is provided by Gutenberg in https://github.com/WordPress/gutenberg/pull/74075. We should just be able to send events from the provider and remove the local modal.
2. Provider handling of unrecoverable errors - we still just try reconnecting regardless of the reason.

### DIFF
--- a/src/websocket-client.ts
+++ b/src/websocket-client.ts
@@ -239,12 +239,26 @@ export function createWebSocketConnection( serverUrl: string ): ProviderCreator 
 
 			// Provide some debugging functions in development mode.
 			if ( isDevelopment() ) {
+				// Because there are multiple providers, disconnectWebSocket() and
+				// reconnectWebSocket() will be overridden by the last entity or collection
+				// provider created. Call the previous function if present.
+
+				const previousDisconnectFunction = window.VIP_RTC.debug.disconnectWebSocket;
 				window.VIP_RTC.debug.disconnectWebSocket = () => {
+					if ( previousDisconnectFunction ) {
+						previousDisconnectFunction();
+					}
+
 					provider.off( 'connection-close', connect );
 					provider.disconnect();
 				};
 
+				const previousReconnectFunction = window.VIP_RTC.debug.reconnectWebSocket;
 				window.VIP_RTC.debug.reconnectWebSocket = () => {
+					if ( previousReconnectFunction ) {
+						previousReconnectFunction();
+					}
+
 					provider.on( 'connection-close', connect );
 					void connect();
 				};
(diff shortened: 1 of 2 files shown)


src/websocket-client.ts
```diff
@@ -227,6 +256,14 @@ export function createWebSocketConnection( serverUrl: string ): ProviderCreator
...
 			return {
 				destroy: () => provider.destroy(),
+				on: < K extends keyof ProviderEventMap >(
+					event: K,
+					callback: ( data: ProviderEventMap[ K ] ) => void
+				) => {
+					if ( event === 'status' ) {
+						statusListeners.push( callback as ( state: SyncConnectionState ) => void );
+					}
+				},
```
chriszarate: We expected this to be a simple pass-through to `provider.on()`. does that not work? If not, it might need upstream work.


```suggestion
				on: < K extends keyof ProviderEventMap >(
					event: K,
					callback: ( data: ProviderEventMap[ K ] ) => void
				) => provider.on( event, callback ),
```

alecgeatches: This should probably work! Although this change as-is causes some type errors.

src/websocket-client.ts
```diff
@@ -197,27 +207,44 @@ export function createWebSocketConnection( serverUrl: string ): ProviderCreator
...
+			provider.on( 'status', event => {
+				onStatusChange( objectType, objectId, event );
 			} );
-			provider.on( 'status', event => onStatusChange( objectType, objectId, event ) );
 
 			// Provide some debugging functions in development mode.
 			if ( isDevelopment() ) {
+				// Because there are multiple providers, disconnectWebSocket() and
+				// reconnectWebSocket() will be overridden by the last entity or collection
+				// provider created. Call the previous function if present.
```
chriszarate, line 230: do we care?

alecgeatches: In testing, `VIP_RTC.debug.disconnectWebSocket()` only affected the last provider that was set up, as the window function is overwritten by each subsequent entity setup. In my case, this was usually the `category-collection` websocket connection, which doesn't show anything if you disconnect it. This functionally made the disconnect/reconnect functions do nothing. If we don't care about ever using the debug functions we can remove it, but this will correctly disconnect all providers.

chriszarate: Ok, we can work towards a better debug approach for this separately.

chriszarate (approved):

PR #148, opened 2026-01-27
Update awareness to work with trunk

ingeniumed:
### Description

This is the companion PR for https://github.com/WordPress/gutenberg/pull/74962.

The awareness code has been changed a lot in trunk, and as a result our implementation is broken. This changes the way it works to use what Gutenberg provides. As a result, notifications and debug data export don't work and that's been removed. I have had to make some type changes that ideally, if the node packages used were updated would just work as expected. So when that happens, we can hopefully drop a lot of the types I have hardcoded.

I have also commented out the `setConnectionStatus` code as that's gone. @shekharnwagh has a PR to resolve it, which is why I haven't bothered to fix it.


src/components/avatar.tsx
```diff
@@ -24,8 +24,11 @@ export function Avatar( {
 		.filter( Boolean )
 		.join( ' ' );
 
+	const avatarUrl =
```
ingeniumed, line 27: avatar_url doesn't exist on the user, whcih is why its done this way

src/components/avatars.tsx
```diff
@@ -1,23 +1,25 @@
...
 import { Avatar } from '@/components/avatar';
 import { CollaboratorsList } from '@/components/collaborators-list';
-import { useActiveUsers } from '@/hooks/use-post-editor-awareness-state';
 import { type CursorRegistry } from '@/utilities/cursor-registry';
 
 import '@/components/avatars.scss';
 
 interface AvatarsProps {
 	cursorRegistry: CursorRegistry;
+	postId: number | null;
```
ingeniumed, line 13: I chose to pass this rather than constantly get it in every component

src/components/rtc-settings-panel.tsx
```diff
@@ -26,9 +26,6 @@ interface SelectResult {
 	isAvatarsEnabled: boolean;
 	isCursorsEnabled: boolean;
 	isDebugToolsEnabled: boolean;
-	isPostUpdateNotificationEnabled: boolean;
```
ingeniumed, line 29: As mentioned, notifications aren't supported at the moment so this is gone

src/components/rtc-settings-panel.tsx
```diff
@@ -125,46 +128,9 @@ export function RTCSettingsPanel() {
...
-						checked={ isUserExitNotificationEnabled }
-						onChange={ ( enabled: boolean ) =>
-							setSetting( Setting.USER_EXIT_NOTIFICATION, enabled )
-						}
-					/>
-
 					<Heading level={ 3 } style={ { marginTop: '24px' } }>
 						{ __( 'Debug', 'vip-real-time-collaboration' ) }
 					</Heading>
-					<DebugDataExportButton />
```
ingeniumed, line 167: I would have kept this but the function it uses not existing and the types failing kept causing failures. So I removed this. I have to implement this functionality in a better way in trunk

types/wordpress__core-data/index.d.ts
```diff
@@ -1,12 +1,92 @@
...
+	export const store: {
 		name: string;
 	};
 
-	interface CoreDataSelectors {
+	export interface CoreDataSelectors {
 		getCurrentUser(): User;
 	}
+
+	// Types from awareness/types.ts
```
ingeniumed, line 15: Hopefully once the packages are updated, this won't be needed as the published versions will have the right types

types/wordpress__sync/index.d.ts
```diff
@@ -1,53 +1,73 @@
 /**
- * External dependencies
+ * Type declarations for @wordpress/sync.
```
ingeniumed: Same thing as core-data

shekharnwagh:
Haven't looked at the diff yet, but testing this PR I found 2 issues -
1. The styles for RTC settings in sidebar seem off - there is not enough vertical spacing between settings 
    Screenshot 2026-01-27 at 9 06 28 PM

2. When the 'Enable Avatars' setting is toggled off, the avatars disappear. But when its toggled back on, the avatars are not visible. There was an empty div in the markup, but its empty.

    https://github.com/user-attachments/assets/905a1c56-736b-4d8e-9da9-843d5d3c82d4

maxschmeling:
Good finds @shekharnwagh 

> The styles for RTC settings in sidebar seem off - there is not enough vertical spacing between settings

Fixed in https://github.com/Automattic/vip-real-time-collaboration/pull/148/commits/270b3f1a1eb03f12eb13a49f6add279521b77108

> When the 'Enable Avatars' setting is toggled off, the avatars disappear. But when its toggled back on, the avatars are not visible. There was an empty div in the markup, but its empty.

Fixed in https://github.com/Automattic/vip-real-time-collaboration/pull/148/commits/3d9084214ab5dab052cc968b3ead3d475ea20cb0

shekharnwagh:
> Good finds @shekharnwagh
> 
> > The styles for RTC settings in sidebar seem off - there is not enough vertical spacing between settings
> 
> Fixed in [270b3f1](https://github.com/Automattic/vip-real-time-collaboration/commit/270b3f1a1eb03f12eb13a49f6add279521b77108)
> 
> > When the 'Enable Avatars' setting is toggled off, the avatars disappear. But when its toggled back on, the avatars are not visible. There was an empty div in the markup, but its empty.
> 
> Fixed in [3d90842](https://github.com/Automattic/vip-real-time-collaboration/commit/3d9084214ab5dab052cc968b3ead3d475ea20cb0)

Thank you for fixing it ! It seems to work with. A small observation, the bottom margin was `12px` and now its `16px` and the difference is visible at first glance.

Edit: I've pushed a fix

shekharnwagh (approved):

chriszarate (approved):

PR #164, opened 2026-02-16
Update plugin to match Gutenberg changes

chriszarate:
- Update WebSocket provider to use tweaked `onStatusChange` from https://github.com/WordPress/gutenberg/pull/75066
- Remove unused client code now handled by Gutenberg `trunk`, including cursor awareness
- Remove unused server code now handled by Gutenberg `trunk`, including settings


inc/Compatibility/Compatibility.php
```diff
@@ -37,33 +30,6 @@ public static function admin_notices(): void {
 		}
 	}
 
-	/**
-	 * Force-enable sync collaboration experiment.
-	 *
-	 * @psalm-suppress PossiblyUnusedReturnValue Psalm does not detect usage via add_filter.
-	 */
-	public function enable_sync_collaboration_experiment( mixed $experiments ): array {
```
ingeniumed, line 45: Right now, the experiment has to be manually turned on. Shouldn't we update this so that when you install our plugin the experiment is automatically turned on?

inc/Settings/Settings.php
```diff
@@ -12,91 +12,17 @@ final class Settings {
...
-	public const GUTENBERG_OPTION_NAME = 'gutenberg_enable_real_time_collaboration';
+	public const GUTENBERG_OPTION_NAME = 'enable_real_time_collaboration';
 
 	public function __construct() {
-		add_action( 'admin_init', [ $this, 'register_settings' ] );
 		add_action( 'admin_menu', [ $this, 'add_options_page' ] );
-		add_filter( 'pre_option_' . self::GUTENBERG_OPTION_NAME, [ $this, 'filter_gutenberg_rtc_option' ], 10, 0 );
 	}
 
 	public static function is_vip_rtc_enabled(): bool {
```
ingeniumed, line 21: Is this function still needed?

This doesn't actually work as it was originally intended. Disabling the GB option doesn't do anything here.

inc/Settings/Settings.php
```diff
@@ -148,47 +65,13 @@ public static function settings_page_content(): void {
...
-					<?php checked( $value, false ); ?>
-				/>
-				<?php echo esc_html( $args['disabled_label'] ); ?>
-			</label>
-			<p class="description">
-				<?php echo esc_html( $args['description'] ); ?>
+			<p>
+				<?php
+				/* translators: %s: Gutenberg commit hash */
+				echo esc_html( __( 'Real-time collaboration can be disabled on the Writing settings page.', 'vip-real-time-collaboration' ) );
```
ingeniumed: Could we link directly to options-writing.php instead?

inc/Settings/Settings.php
```diff
@@ -148,47 +65,13 @@ public static function settings_page_content(): void {
...
-					value="0"
-					<?php checked( $value, false ); ?>
-				/>
-				<?php echo esc_html( $args['disabled_label'] ); ?>
-			</label>
-			<p class="description">
-				<?php echo esc_html( $args['description'] ); ?>
+			<p>
+				<?php
+				/* translators: %s: Gutenberg commit hash */
```
ingeniumed: Not needed?

inc/Settings/Settings.php
chriszarate: it does, but the option name changed recently

inc/Compatibility/Compatibility.php
chriszarate: there is no experiment anymore

inc/Settings/Settings.php
chriszarate: probably! good suggestion

inc/Compatibility/Compatibility.php
ingeniumed: I used the word experiment to describe our feature. The point is that should we be turning on the feature in our plugin?

types/wordpress__sync/index.d.ts
```diff
@@ -8,28 +8,30 @@ declare module '@wordpress/sync' {
...
+
+		/**
+		 * Optional error information when status is 'disconnected'.
+		 */
+		error?: ConnectionError;
 	}
 
 	interface ProviderEventMap {
-		'sync-connection-status': SyncConnectionState;
+		status: SyncConnectionState;
```
shekharnwagh:
```suggestion
		status: ConnectionStatus;
```

src/components/rtc-settings-panel.tsx
shekharnwagh, line 1: We are no longer registering this panel, so should this be removed ?

inc/Compatibility/Compatibility.php
shekharnwagh: Yeah, when I was testing - it was turned off by default which would mean collaboration would stop working for VIP customers until they turn it back on.

src/components/rtc-settings-panel.tsx
chriszarate: I left a few unused things in case they'd be useful again:

- Notifications code (might be ported)
- Debug tools
- Settings panel (for debug tools, if we can restore them)

shekharnwagh (approved):

src/components/rtc-settings-panel.tsx
shekharnwagh: Do you think adding a comment at top of the file might be useful so that its clear ?

inc/Compatibility/Compatibility.php
chriszarate: Fixed in https://github.com/Automattic/vip-real-time-collaboration/pull/164/changes/8d75e26bbc4789bd2818dbc3a4e1fde1b3645be1

inc/Settings/Settings.php
chriszarate: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/164/commits/e07de76e9c8b02bdb7bf79501e34af867f26f9c1 and https://github.com/Automattic/vip-real-time-collaboration/pull/164/commits/3e2194ce9d2e78d70be1354f3d63f53c9e217242

ingeniumed:
Opened https://github.com/Automattic/vip-real-time-collaboration/pull/168 after this is merged in.

ingeniumed (approved):

PR #182, opened 2026-04-14
v0.2.6

shekharnwagh:


shekharnwagh:
I looked over commist since last release, and it appears that they are all related to the documentation. Thus, we are keeping this open until we have changes that warrant a new release.

PR #184, opened 2026-04-20
[VIP-3049] Add collaborator limits to websocket server

shekharnwagh:
## Summary

Adds per-environment collaborator and connection limits for real-time collaboration. The websocket server rejects new users when limits are hit; the editor shows a custom modal explaining the state. For generic disconnects, Gutenberg's default modal now renders a retry countdown.

## WebSocket server

- New env var `VIP_RTC_WS_MAX_COLLABORATORS` (default `300`; `-1` for unlimited). Counts unique WP `user_id` from the JWT, so one user across multiple tabs counts once.
- New close code `4003` for collaborator-limit rejections. The existing connection-capacity cap (close code `4002`) is unchanged.
- New `wpvip_rtc_active_collaborators` gauge for monitoring.
- Tightened JWT payload validation; a `null` `user_id` fails closed at the limit check.
- `JWT_SECRET` missing-env guard moved from `config.ts` to the server entry point so tests can import `config` without calling `process.exit`.

## Editor

- Map close codes `4002` and `4003` to `connection-limit-exceeded` and `collaborator-limit-exceeded` sync error codes.
- Custom modal for both limit errors with error-specific copy, a post-type-aware "Back to Posts" action, and a "Copy Post Content" escape hatch for unsaved work. Gutenberg's default modal is suppressed for these codes via the `editor.isSyncConnectionErrorHandled` filter.
- The "contact your WordPress VIP Relationship Manager" line is shown only to users with `manage_options`. Capabilities are passed to the client via the `VIP_RTC` inline script.
- For non-limit disconnects, emit `willAutoRetryInMs` and `backgroundRetriesFailed` so Gutenberg's default modal renders a countdown and only reveals after ~3 consecutive failures instead of flashing on every blip.
- Fix: first retry after a live session drops now waits the emitted 1s backoff instead of retrying with zero delay.
- Fix: gate the retry-attempts reset on 2s of stable connection. y-websocket emits `status: 'connected'` on raw WebSocket open, before server-side 4002/4003 rejections arrive, which was resetting attempts and stalling reconnect in a ~1s loop.

## Testing

Start the local environment with `npm run dev` (WordPress at `http://localhost:8888`; websocket server is started alongside).

Manual flows:

- **Collaborator limit (4003)**: set `VIP_RTC_WS_MAX_COLLABORATORS=1` on the websocket server, open a post as user A, then open the same post as user B. User B sees the "Collaborator limit reached" modal; user A keeps editing.
- **Connection limit (4002)**: set `VIP_RTC_WS_MAX_CONNECTIONS` low, open enough tabs to exceed it. The next tab sees the "Connection limit reached" modal.
- **Default modal with countdown**: with the editor open, stop the websocket server. After ~6s of retries Gutenberg's default disconnect modal appears with a live countdown; bringing the server back up auto-closes it.
- **Copy Post Content**: trigger either limit modal, make local edits, and click "Copy Post Content" to salvage them before navigating away.

Unit tests for the websocket server:

```
cd websocket-server && npm test
```


src/components/collaboration-limit-modal.tsx
```diff
@@ -0,0 +1,123 @@
...
+		};
+	}
+
+	return {
+		title: __( 'Connection limit reached', 'vip-real-time-collaboration' ),
+		body: __(
+			"The collaboration server for this environment is at capacity. We'll keep trying to reconnect automatically. To raise the limit, contact your Relationship Manager.",
+			'vip-real-time-collaboration'
+		),
+	};
```
shekharnwagh, line 62: @smithjw1 The text displayed in the modal when the connection and collaborator limits are reached. Does this seem good, or do you think it needs to be changed?

smithjw1: Please change this sentence:

> To raise the limit, contact your Relationship Manager.

Can we make it conditional based on the user role? If not, let's just leave it out, since many editorial users won't know they even have a Relationship Manager.

If we can filter by user role, if the user is an Administrator, it should say:

> To change this limit, contact your WordPress VIP Relationship Manager.

shekharnwagh: Updated in f09f2f0

maxschmeling (approved):

PR #186, opened 2026-05-15
Add custom messaging for collaborator limit reached

smithjw1:
This pull request introduces a new "collaborator limit" dialog that detects when the maximum number of real-time collaborators has been reached, displays a user-facing dialog in the editor, and provides a workflow for users to request an upgrade. 

It also adds backend support for telemetry tracking when the dialog is shown, exposes relevant configuration to the frontend, and includes comprehensive tests for the new API endpoint.

**Collaborator Limit Dialog & Upgrade Workflow:**

- Adds a React-based `CollaboratorLimitDialog` component that appears in the editor when the real-time collaborator limit is exceeded, informs users of the limit, and provides an upgrade path via either a support ticket or mailto fallback.
- Registers the dialog as a plugin in the editor and ensures the error is handled gracefully, preventing other error handling from interfering.
- Implements the `submitUpgradeTicket` utility, which attempts to submit a support request via AJAX and falls back to a mailto link if necessary, with full test coverage. [[1]](diffhunk://#diff-bbb0a17af9f5d979b0ca914bd03ef545a1a9e2dd3da03658ae40ee97a3894261R1-R103) [[2]](diffhunk://#diff-e01ced8038bc9880647f7399945febe944bac10993e8e57896f3b2c7e3551574R1-R121)

**Backend API & Telemetry:**

- Introduces a new REST API controller, `TelemetryApiController`, with a `/telemetry/limit-dialog` endpoint to record when the collaborator limit dialog is shown, firing a custom action for telemetry.
- Registers the new telemetry API endpoint in the main REST API initialization.
- Adds integration tests to verify the telemetry endpoint fires the correct action and enforces authentication.

**Configuration & Asset Loading:**

- Exposes new configuration values (`collaboratorLimit`, `collaboratorLimitTier`, `contactAjax`, `supportEmail`) to the frontend via the asset loader, with backend methods and filters for customization. [[1]](diffhunk://#diff-02f018b02960846ed2e34995f29e8ffb48b309b8576e427c6c349fe59912d1e0R67-R70) [[2]](diffhunk://#diff-02f018b02960846ed2e34995f29e8ffb48b309b8576e427c6c349fe59912d1e0R84-R121) [[3]](diffhunk://#diff-d98040c386fddf530d3e37ba6c2f465673e31a3f4a1bb53b54f4903f0e08c390R6-R10)
- Updates the frontend config utility to read these new values.

**Dependency Updates:**

- Adds required WordPress packages for the new dialog and state management functionality.

**Editor Integration:**

- Wires up the collaborator limit dialog registration in the main editor entrypoint to ensure the feature is always available.


shekharnwagh (changes requested):
Thanks for working on this !
The upgrade-ticket flow and telemetry endpoint look good. But there are a few issues that need addressing before this can ship:

**1. This duplicates the existing limit modal.** Trunk already has [`CollaborationLimitModal`](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/src/components/collaboration-limit-modal.tsx), registered in [`src/index.ts`](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/src/index.ts#L25-L36), handling `connection-limit-exceeded` and `collaborator-limit-exceeded`. With this PR, both non-dismissible modals render stacked when the error fires. Could we fold the upgrade CTA into the existing modal instead of adding a parallel one? That also makes the duplicate `editor.isSyncConnectionErrorHandled` filter in `register.tsx` unnecessary.

**2. Wrong error code.** The dialog targets `connection-limit-exceeded` (close code 4002 = server connection capacity), but the collaborator-limit copy matches `collaborator-limit-exceeded` (4003). See [`getErrorFromCloseCode` in `src/websocket-client.ts`](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/src/websocket-client.ts#L93-L108).

**3. `getSyncConnectionStatus` is private in current Gutenberg** ([WordPress/gutenberg#78097](https://github.com/WordPress/gutenberg/pull/78097), now in [`core-data/src/private-selectors.ts`](https://github.com/WordPress/gutenberg/blob/trunk/packages/core-data/src/private-selectors.ts)). Reading it from the public `select( 'core' )` store returns `undefined`, so the dialog never shows. Please use the existing [`getCoreDataSelectors()` helper](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/src/utilities/sync-connection-status.ts), which the existing modal already uses.

**4. The displayed limit isn't connected to the enforced limit.** Nothing in mu-plugins implements `vip_rtc_collaborator_limit`/`vip_rtc_collaborator_limit_tier`, so every site will show "10" / "Standard" while the server enforces [`VIP_RTC_WS_MAX_COLLABORATORS`](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/websocket-server/config.ts#L22) (default 1, enforcement added in [#184](https://github.com/Automattic/vip-real-time-collaboration/pull/184)). We either need platform-side wiring for these filters before merge, or the server should communicate its limit to the client so the copy can't drift.

**5. Hardcoded `/wp-admin/edit.php`** in `navigateToPostsList` breaks subdirectory installs. The existing modal's relative [`edit.php`](https://github.com/Automattic/vip-real-time-collaboration/blob/trunk/src/components/collaboration-limit-modal.tsx#L103) handles this.

Smaller items:
- `types/wordpress__data__lock-unlock/index.d.ts` is unused. Leftover from an earlier approach?
- `permissions_check` could be `current_user_can( 'edit_posts' )` since only editor users can trigger the dialog.
- `supportEmail` is hardcoded while its sibling values are filterable.

smithjw1:
Thanks for the thorough review, @shekharnwagh — all addressed in 3311b52. The core change is folding the upgrade CTA into the existing `CollaborationLimitModal` rather than shipping a parallel dialog, which resolves most points at once.

**1. Duplicate modal.** Folded the Upgrade CTA, telemetry, and a success view into `CollaborationLimitModal`. Deleted `CollaboratorLimitDialog`, its `register.tsx`, and the redundant `editor.isSyncConnectionErrorHandled` filter — the existing filter in `src/index.ts` already covers both codes. No more stacked modals.

**2. Wrong error code.** The CTA + telemetry now fire only for `collaborator-limit-exceeded` (4003). The transient `connection-limit-exceeded` (4002) keeps its existing Back/Copy buttons since an upgrade wouldn't help there.

**3. Private `getSyncConnectionStatus`.** Fixed for free by folding in — the modal already reads via `getCoreDataSelectors()`.

**4. Displayed limit vs. enforced limit.** Dropped the hardcoded number/tier entirely (copy stays generic, like the existing modal) and removed `collaboratorLimit`/`collaboratorLimitTier` from the config, types, and ticket payload. Nothing can drift from `VIP_RTC_WS_MAX_COLLABORATORS`, and there's no platform-wiring dependency to block merge. If we later want to surface the actual number, the cleanest path is having the server communicate it to the client — happy to follow up separately.

**5. Hardcoded `/wp-admin/edit.php`.** Gone — now reuses the existing modal's relative `edit.php` navigation, so subdirectory installs work.

Smaller items:
- Removed the unused `types/wordpress__data__lock-unlock/index.d.ts`.
- `permissions_check` now uses `current_user_can( 'edit_posts' )`; tests updated to cover editor (pass) / subscriber + anonymous (403).
- `supportEmail` is now filterable via `vip_rtc_support_email`, matching its siblings.

Verified: tsc, eslint, prettier, phpcs, psalm clean; JS unit tests (13) and PHP integration tests (14) all pass.

shekharnwagh:
@smithjw1 With the new changes, I was able to test it locally. Clicking "Upgrade Plan" opened the default mail app with a pre-filled draft, which is the expected mailto fallback since it was running outside a VIP environment. A few smaller things I noticed while going through the updated diff:

1. `getSite()` needs `manage_options`, so site data never loads for most users. `getSite` is backed by [`/wp/v2/settings`](https://github.com/WordPress/gutenberg/blob/trunk/packages/core-data/src/entities.js#L506-L511), which only admins can read. Editors and authors get a background 403 and `site` stays null, so the fallbacks always kick in.

2. `userEmail` will always be empty. [`getCurrentUser` resolves `/wp/v2/users/me` in view context](https://github.com/WordPress/gutenberg/blob/trunk/packages/core-data/src/resolvers.js#L54-L57), and `email` is an edit-context-only field. The ticket path is fine because the handler falls back to the server-side user email, but the mailto body ends up as `Requested by: Alice <>`.

3. `upgradeState` never resets. If a user submits, the connection recovers, and the limit is hit again later in the same session, the modal reopens already showing "Request sent" with no error context or Upgrade button. What do you think about resetting `upgradeState` to `idle` in the existing visibility effect when the status becomes `connected`, so each limit episode starts fresh?

4. The success view drops "Copy Post Content". After submitting, the user is still disconnected and any unsaved work is still trapped in the editor, so the copy button might be useful. What do you think of keeping it in the success view, or rendering the success content inside the single existing modal and only varying the body text and buttons ?

shekharnwagh:
Another open question: Gutenberg defaults to showing the modal after the 3rd failure for other cases to give it time to recover from transient issues, but for connection and collaborator limits, it appears immediately. I can see not doing it because connection and collaborator limits won't resolve until a connected collaborator disconnects.

smithjw1:
> Another open question: Gutenberg defaults to showing the modal after the 3rd failure for other cases to give it time to recover from transient issues, but for connection and collaborator limits, it appears immediately. I can see not doing it because connection and collaborator limits won't resolve until a connected collaborator disconnects.

I think I agree. I think we should show this message right away, not wait for three failures, since the cause of the failure is less transient.

smithjw1:
@shekharnwagh thanks for walking me through this! I've made those updates.

shekharnwagh (approved):
Thank you for the quick updates to address the feedback. LGTM 🚢

PR #199, opened 2026-06-16
Add client-side per-room connection limit

smithjw1:
## Summary

Adds an optional, client-side cap on the number of concurrent **connections** (browser tabs/clients) to a single sync room, honoring the `sync.pollingProvider.maxClientsPerRoom` filter — the same hook Gutenberg's default polling transport exposes — so sites that already cap the default transport keep working after the VIP WebSocket provider takes over.

## Behavior

- **No-op unless opted in.** `setupRoomClientLimit` bails immediately when `getMaxClientsPerRoom` returns a non-positive / non-finite value (the default with no filter). With no enforcing filter, the connect / reconnect / modal flow is byte-for-byte identical to `trunk` — only an extra side-effect-free `applyFilters` call per room runs.
- **When enforced**, each client ranks the room's connections by join time (read from Gutenberg's own `collaboratorInfo.enteredAt`, `clientId` tie-break) and the **newest** connections past the cap disconnect themselves. Counting is **per connection (per tab)**, not per unique user — matching the filter name and Gutenberg's own `checkConnectionLimit` semantics.
- The yield is **terminal**: it surfaces a dedicated `room-limit-exceeded` modal and does **not** reconnect, avoiding a reconnect/evict thrash that would otherwise trip the server's connection limit (observed as a follow-on `4002`).

## Implementation notes

- Reads Gutenberg's existing `collaboratorInfo` awareness field rather than writing a custom field (Gutenberg's `AwarenessWithEqualityChecks` throws on unregistered fields).
- Dedicated `room-limit-exceeded` error code + modal copy (not reused from the server-side collaborator/connection limits, whose "we'll keep trying" copy is wrong for this terminal path).
- Re-evaluates on every `connected` status, not only on awareness changes, so an over-limit client is caught even when awareness was already present before the socket opened.

## Scope / safety

This is **best-effort, cooperative** enforcement (simultaneous joins can briefly exceed the cap; a modified client can ignore it). Authoritative per-room limits must live on the server — tracked in **VIP-3150**.

## Testing

- Unit tests for the ranking/parsing helpers (`room-client-limit.test.ts`).
- Manually verified in `wp-env` with a throwaway mu-plugin capping post 2: latecomer/extra-tab yields with the modal, established clients keep editing, no reconnect thrash.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/constants/sync-errors.ts
+++ b/src/constants/sync-errors.ts
@@ -8,7 +8,7 @@ import type { ConnectionErrorCode } from '@wordpress/sync';
 export const CUSTOM_MODAL_ERROR_CODES: ReadonlyArray< ConnectionErrorCode > = [
 	'collaborator-limit-exceeded',
 	'connection-limit-exceeded',
-	'room-limit-exceeded',
+	'room-connection-limit-exceeded',
 ];
 
 /**
--- a/src/constants/sync-errors.ts
+++ b/src/constants/sync-errors.ts
@@ -8,6 +8,7 @@ import type { ConnectionErrorCode } from '@wordpress/sync';
 export const CUSTOM_MODAL_ERROR_CODES: ReadonlyArray< ConnectionErrorCode > = [
 	'collaborator-limit-exceeded',
 	'connection-limit-exceeded',
+	'room-limit-exceeded',
 ];
 
 /**
--- a/src/components/collaboration-limit-modal.tsx
+++ b/src/components/collaboration-limit-modal.tsx
@@ -33,6 +33,18 @@ function getConnectionErrorMessage(
 	errorCode: ConnectionErrorCode | undefined,
 	isAdmin: boolean
 ): ConnectionErrorMessage {
+	if ( errorCode === 'room-limit-exceeded' ) {
+		// Client-side per-room connection cap. This yield is terminal — the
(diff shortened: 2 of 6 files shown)


src/constants/sync-errors.ts
```diff
@@ -8,6 +8,7 @@ import type { ConnectionErrorCode } from '@wordpress/sync';
 export const CUSTOM_MODAL_ERROR_CODES: ReadonlyArray< ConnectionErrorCode > = [
 	'collaborator-limit-exceeded',
 	'connection-limit-exceeded',
+	'room-limit-exceeded',
```
maxschmeling: `room-limit-exceeded` sounds like a client has connected to too many rooms. Should we call it `collaborator-limit-exceeded` or something?

maxschmeling (approved):

src/constants/sync-errors.ts
smithjw1: Good call — renamed to `room-connection-limit-exceeded` in 6f27049 to make clear it's a per-room connection (tab) cap, and to keep it distinct from `collaborator-limit-exceeded` (the env-wide collaborator limit) and `connection-limit-exceeded` (server capacity).

PR #208, opened 2026-07-22
Add multiplexed WebSocket protocol

shekharnwagh:
Closes VIP-3242

The editor currently opens one WebSocket per collaboration room, so a session syncing several entities holds several connections at once. This PR adds multiplexing support: all rooms in a browser session share one physical WebSocket, and a small versioned protocol routes each room's traffic over it.

How it works:

- A small versioned binary protocol (`vip-rtc-multiplex-v1`) carries all room traffic over the shared socket; the frame diagram and table below show the messages.
- Every room joins the same way. The grant in the upgrade URL only authenticates the connection itself; each room, including the first, joins by sending a `subscribe` message with its own grant, which the server checks before sending any room data.
- Repeating `subscribe` for an active room is idempotent: the server keeps the existing room adapter and sends another acknowledgement.
- On the client, a `WebSocketPolyfill` adapter gives y-websocket a virtual socket per room, so existing provider behavior (offline recovery, room limits, 4001 token rotation) keeps working without changes to the sync engine.
- Multiplexing is off by default, behind the `VIP_RTC_WS_MULTIPLEXING_ENABLED` PHP constant. With the flag off, the client keeps the legacy one-socket-per-room transport. Integration Center support for the flag needs follow-up changes in vip-go-api and vip-go-mu-plugins.
- The legacy room-URL transport is still available on the server, so the multiplex-capable server can be deployed before the plugin.

### The messages

Here is a real frame, byte by byte. Every message type starts with the same prefix (version, type, room name) and swaps the type-specific part at the end:

x_subscribe_frame

What each type means and who may send it:

| Message       | Sent by | Meaning                                                                          |
| ------------- | ------- | -------------------------------------------------------------------------------- |
| `subscribe`   | client  | Ask to join a room; carries a grant that authorizes this room only               |
| `subscribed`  | server  | The room is joined; room data can now flow                                       |
| `data`        | both    | An unchanged y-websocket message, wrapped with its room                          |
| `unsubscribe` | client  | Leave a room; the connection stays open                                          |
| `room_closed` | server  | The server closed one room; carries the code (`4004` terminal, `4005` retryable) |

Each frame belongs to exactly one room, so the room name in the prefix is the routing key: both sides look it up and hand the payload only to that room's socket and Yjs doc. Direction is enforced: if a client sends a server-only message, the server closes the connection as a protocol error.

### The flow

A session with two rooms, where one is rejected and the other keeps working:

```mermaid
sequenceDiagram
    participant E as Editor (rooms A and B)
    participant S as Server
    E->>S: upgrade ?auth=grant
    Note over E,S: connection open, no rooms yet
    E->>S: subscribe(A, grantA)
    S->>E: subscribed(A)
    S->>E: data(A, ...)
    E->>S: subscribe(B, grantB)
    S->>E: room_closed(B, 4004)
    Note over E,S: B is rejected, A keeps syncing
    S->>E: data(A, ...)
```

### Shared code

The plugin and the WebSocket server share one module: `websocket-server/protocol.ts`. The client code in `src/shared-websocket.ts` imports the encode/decode functions, the message types, and the subprotocol name from it. The module has no Node-only code, so it bundles fine for the browser.

Sharing it means both sides always agree on the wire format. A change to the protocol is one edit in one file, and the client and server both pick it up. The client tests also use the same codec to build real server frames.

### New dependencies

- `lib0` (runtime, plugin and server): the protocol codec is built on lib0's encode and decode helpers for varuints, strings, and byte arrays. Yjs already uses lib0 internally, so nothing new is being pulled in.
- `ws` (dev only, root package): the end-to-end test uses it to connect to the server as a plain WebSocket client from Node. It doesn't ship with the plugin. Direct and transitive `ws` packages are pinned to patched versions.

### Partial room failures

Before this change, every room had its own physical WebSocket, so a room failure was isolated automatically. With a shared connection, one bad room must not take down the healthy ones, so this PR keeps that isolation explicitly:

- A terminal room rejection (`4004`) or a room-limit yield closes only that virtual room. The other subscribed rooms stay connected.
- A retryable room interruption (`4005`, for example a grant that expired between connecting and subscribing) fetches a fresh grant for just that room and re-subscribes it with the usual backoff.
- Connection-scoped closes still affect every room. Retryable ones like 4001 restore all active rooms over one replacement socket, while protocol and policy failures stay terminal.
- The error UI is unchanged: a failed room shows the same editor modals as before. Room-aware messaging is a possible follow-up, not part of this PR.

### Cleanup and observability

- The physical heartbeat runs every 30 seconds. If an open connection has no active rooms at a heartbeat tick, the server records `wpvip_rtc_connection_failures_total{reason="zero_room_timeout"}` and closes it with `4001`. Active sessions retain the separate four-hour `4001` rotation.
- Before terminating an unresponsive physical socket, the server closes every room adapter. This removes its Yjs connection state and clears its room heartbeat interval without waiting for the physical `close` event.
- `wpvip_rtc_active_room_connections` reports current logical rooms, `wpvip_rtc_peak_rooms_per_connection` records the peak for each completed physical connection, and `wpvip_rtc_room_connections_closed_total{reason}` uses `authorization_rejected`, `grant_expired`, `client_unsubscribe`, `server_room_close`, or `physical_connection_close`. These metrics do not include room or identity labels.
- Errors from `ws` rejecting an invalid frame are handled per connection instead of becoming uncaught process errors.

### Review notes

- The commits are self-contained and easiest to review in order: they build up protocol, server, client, and tests, then harden and instrument the result.
- The architecture doc gains a short section describing the transport and the feature flag, reusing the frame diagram above.

## Testing

- `npm run test:js:unit`
- `npm --prefix websocket-server test`
- `npm run lint:js`
- `tests/e2e/specs/multiplexed-rtc.spec.ts` (3 passed locally)
- Test the usual collaboration flows to make sure nothing regressed: two clients editing the same post, going offline and reconnecting, and token rotation.


chriszarate:
> Room-aware messaging is a possible follow-up, not part of this PR.

Understanding that this can come later, can you describe how you expect it to work?

> A physical connection with zero subscribed rooms is a legal state (between the upgrade and the first subscribe).

Is there a timeout that will eventually close the connection if there continue to be no subscribed rooms?

shekharnwagh:
> > Room-aware messaging is a possible follow-up, not part of this PR.
> 
> Understanding that this can come later, can you describe how you expect it to work?

Room failures are isolated at the transport layer, and Gutenberg receives a separate provider error for the affected room. The client already knows and logs the affected room, close code, and retry decision. A follow-up would pass that context to the editor UI and use it to customize the modal messaging.

> > A physical connection with zero subscribed rooms is a legal state (between the upgrade and the first subscribe).
> 
> Is there a timeout that will eventually close the connection if there continue to be no subscribed rooms?

The server authenticates the physical connection before the client subscribes, so every room uses the same flow. A subscription normally follows immediately, so the connection should rarely stay empty for long. The four-hour timeout is the only explicit cleanup for a responsive empty connection. I did not add special handling because this case should be rare. Do you think a shorter timeout for this or a metric to track this would be worth adding?

ingeniumed:
> Do you think a shorter timeout for this or a metric to track this would be worth adding?

I def think a metric to track this would be worth adding along with some special handling.

Is there a downside to shortening the timeout? We could use the metric to see if the timeout should be raised.

Thinking a shorter timeout is better than a larger one to ensure connections are freed up quickly in the rare instance that the no subbed rooms problem appears.

ingeniumed:
I’ve completed a first pass and plan to do another as this is a big one.

I had three Qs:

1. Is multiplex-session cleanup dependent on the physical socket’s `close` event when the heartbeat terminates it? If the heartbeat stops making progress, or termination does not promptly produce that event, what eventually cleans up the subscribed rooms?

2. Does each subscribed virtual room still get its own heartbeat interval from `setupWSConnection`? If so, how does that scale as the number of rooms grows, and are those intervals removed across every room cleanup and failure path?

3. Do you think more failure-path metrics would be useful here? As you noted, with multiplexing the isolation previously provided by separate physical sockets is no longer automatic and must be preserved explicitly. The failure paths are where that isolation is most likely to break down, so it would be helpful to detect transport failures, room-level failures, retries, and incomplete cleanup proactively.

shekharnwagh:
> I’ve completed a first pass and plan to do another as this is a big one.

Thank you @ingeniumed for taking a look!

> I def think a metric to track this would be worth adding along with some special handling.
> Is there a downside to shortening the timeout? We could use the metric to see if the timeout should be raised.

I ended up using the heartbeat (every 30s) for the physical WebSocket connection to clean the connection up when there are no rooms using it and record the metric `wpvip_rtc_connection_failures_total{reason="zero_room_timeout"}`. Added in d5f4123

> Is multiplex-session cleanup dependent on the physical socket’s close event when the heartbeat terminates it?

No. In d5f4123, the heartbeat now closes all active room adapters before it terminates an unresponsive physical socket.

> Does each subscribed virtual room still get its own heartbeat interval from setupWSConnection? If so, how does that scale as the number of rooms grows, and are those intervals removed across every room cleanup and failure path?

Yes. Each room retains y-websocket’s 30-second interval, so the timer work grows linearly with the number of rooms. The room adapter answers the ping locally, so it causes no additional network traffic. Closing an adapter clears its interval, and d5f4123 ensures this happens before physical termination.

> Do you think more failure-path metrics would be useful here?

Yes. In 3ecc772 added active-room and peak-rooms-per-connection metrics. In 3b9939f added `wpvip_rtc_room_connections_closed_total{reason}` to distinguish authorization failures, expired grants, client unsubscribes, server room closes, and physical connection closes. Existing metrics continue to cover physical failures, closes, and reconnection time.

shekharnwagh:
Updated the PR description to reflect new commits - d5f4123, 3ecc772, f319316 and 3b9939f

chriszarate:
@shekharnwagh Can we put multiplexing behind a config var (keyed to a constant) so that it can be more easily tested? Edit for clarity: On the client side.

package.json
```diff
@@ -77,9 +77,11 @@
     "prettier": "npm:wp-prettier@3.0.3",
     "ts-loader": "9.5.2",
     "typescript": "5.9.2",
+    "ws": "8.18.2",
```
ingeniumed: Would update this version before it goes out the door. Codex flagged that 8.21.0 fixes https://github.com/advisories/GHSA-96hv-2xvq-fx4p.

ingeniumed:
Did my second pass, and I think there's a noisy neighbour trade off that we may now run into.

Since each room is now a virtual WebSocket, updates from every room pass through a single physical connection. In a busy editor session, one active room could saturate that connection and create a backlog that delays updates for the other rooms sharing it. The heartbeat may not detect that anything's wrong because both the rooms and the physical connection are "working".

It'd be worth thinking about queue management or any other techniques for tackling this.

package.json
shekharnwagh: Updated in 2fb921c

shekharnwagh:
> Can we put multiplexing behind a config var (keyed to a constant) so that it can be more easily tested? Edit for clarity: On the client side.

In 033e16d added a feature flag - `VIP_RTC_WS_MULTIPLEXING_ENABLED` which only enables multiplexing if it's explicitly set as `true`.

> Since each room is now a virtual WebSocket, updates from every room pass through a single physical connection. In a busy editor session, one active room could saturate that connection and create a backlog that delays updates for the other rooms sharing it.

Agreed, it's worth watching, though in practice steady-state Yjs updates are tiny, and large messages are mostly limited to initial sync at page load. Since this PR is already large and the rollout can be managed via the feature flag, it seems like less of a risk at the moment. I'd prefer to observe whether this shows up in practice. If it does, we can add a metric first (e.g. a `bufferedAmount` high-water gauge to spot backlog) and look into queue management as a follow-up after that.

> The heartbeat may not detect that anything's wrong because both the rooms and the physical connection are "working".

Yeah, a delayed-yet-alive connection would look healthy, since the heartbeat only catches a stall severe enough to hold up a pong for 30s or more, so moderate backlog would go unnoticed. Though it'd be observable in practice way before it reaches something which causes that long of a delay.

chriszarate (approved):

ingeniumed:
> I'd prefer to observe whether this shows up in practice

Good with me 👍🏾

ingeniumed (approved):

PR #210, opened 2026-08-06
Update dependencies to the latest versions

ingeniumed:
## Description

- Update the plugin, example projects, and WebSocket server dependencies to their latest compatible versions.
- Configure Composer to resolve dependencies against the plugin’s minimum supported PHP version, PHP 8.1.
- Update WordPress PHP stubs to 7.0 and adjust the necessary Psalm annotations.
- Replace the hand-maintained `@wordpress/sync` declarations with upstream types and a small plugin-specific type layer for VIP WebSocket error codes.
- Configure `wp-env` to download the latest stable Gutenberg release without specifying a version.

## Testing

- `npm run lint:js`
- `npm run lint:phpcs`
- `npm run test:js:unit`
- `npm run build`
- `composer validate --strict`
- Verified the locked Composer dependencies support PHP 8.1.
- Run `npm run dev` and test in two browser sessions


composer.json
```diff
@@ -18,15 +19,18 @@
...
+    "php-stubs/wordpress-tests-stubs": "^7.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
     "phpunit/phpunit": "^9",
     "psalm/phar": "^6.5",
     "slevomat/coding-standard": "^8.15",
     "yoast/wp-test-utils": "^1.2"
   },
   "config": {
+    "platform": {
+      "php": "8.1.0"
```
shekharnwagh: This pin forces a psalm downgrade: `psalm/phar` 6.8.3+ requires PHP ^8.2, so with the platform pinned to 8.1 composer resolved 6.8.2 (trunk had 6.13.0). 6.8.2 crashes on PHP 8.4 (`Call to undefined function _HumbugBoxd84cf39e5d32\array_any()`), breaking `npm run lint:php` there.

The plugin header declares Requires PHP: 8.2, so maybe this should be `"php": "8.2.0"` ?

shekharnwagh (approved):
Thank you for cleaning up the types! I did a manual test on this and didn't find any regression.

composer.json
ingeniumed: Good catch. This should be 8.2 which in turn should update the dependency. I've also removed the `ignore-platform-req` that was blocking the discovery of this on CI.

Pushed this all up in https://github.com/Automattic/vip-real-time-collaboration/pull/210/commits/d149dbf23146842cc4ec259d3c41b73e43bf9ad1.

ingeniumed: I've noted this in the README in https://github.com/Automattic/vip-real-time-collaboration/pull/210/commits/5c97ba03c2af38f36565df56e4cda7b204ef3797

PR #222, opened 2026-08-12
Use explicit multiplex WebSocket paths

shekharnwagh:
## Description

Multiplex connections currently use the base WebSocket URL, which can fail in environments that require an explicit WebSocket path. This PR moves those connections to a dedicated `/vip-rtc` endpoint and updates both the server and client. The service-level name avoids tying the URL to the multiplex implementation and can remain stable if multiplexing becomes the default transport. Direct `/vip-rtc` and prefixed `/_ws/vip-rtc` URLs are both supported.

## Testing

1. Run the local development environment:

   ```sh
   npm run dev
   ```

2. Open the same post in two separate browser sessions and sign in as different users.
3. Make edits from both sessions and confirm normal collaboration works, including synced changes and collaborator presence.
4. In browser DevTools, open **Network → WS** and confirm the shared connection uses `ws://localhost:1234/vip-rtc?auth=...`.

### DIFF
--- a/websocket-server/server.test.ts
+++ b/websocket-server/server.test.ts
@@ -206,8 +206,8 @@ describe( 'WebSocket transport routing', () => {
 	} );
 
 	for ( const [ path, protocols ] of [
-		[ '/multiplex', [ MULTIPLEX_SUBPROTOCOL ] ],
-		[ '/_ws/multiplex', [ 'vip-rtc-multiplex-v2', MULTIPLEX_SUBPROTOCOL ] ],
+		[ '/vip-rtc', [ MULTIPLEX_SUBPROTOCOL ] ],
+		[ '/_ws/vip-rtc', [ 'vip-rtc-multiplex-v2', MULTIPLEX_SUBPROTOCOL ] ],
 	] as const ) {
 		it( `routes ${ path } through multiplex and selects V1 from ${ protocols.join(
 			', '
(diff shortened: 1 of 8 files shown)


websocket-server/server.test.ts
```diff
@@ -494,7 +504,7 @@ describe( 'WebSocket transport routing', () => {
 			const messagesBefore = await metricValue( 'wpvip_rtc_messages_total' );
 			const baseUrl = await startServer( 100 );
 			const room = `site-7/post-common-${ protocol ? 'multiplex' : 'legacy' }`;
-			const path = protocol ? '/' : `/${ room }`;
+			const path = protocol ? '/multiplex' : `/${ room }`;
```
chriszarate: How about `vip-rtc` instead of multiplex as the eventual target for all requests?

shekharnwagh: I used that to hint at what it does, but open to using `vip-rtc` as eventually that'd be the only way it works.

shekharnwagh: Updated in ef1a619

websocket-server/server.ts
```diff
@@ -85,7 +85,7 @@ export function createRtcServer( options: RtcServerOptions ): RtcServer {
 		const initialGrant = authResult.grant;
 		const pathname = getRequestPathname( request );
 		if ( isMultiplex ) {
-			if ( pathname !== '/' && pathname !== '/_ws' ) {
+			if ( ! [ '/vip-rtc', '/_ws/vip-rtc' ].includes( pathname ) ) {
```
chriszarate: Why do we allow connections at `/vip-rtc`? Can we simplify?

chriszarate (approved):

websocket-server/server.ts
shekharnwagh: It was to make it work with config in `wp-env.json` for local development. I agree, having a single version is better. Updated in cdad45e

PR #225, opened 2026-08-19
Enable Gutenberg’s real-time collaboration experiment by default

ingeniumed:
## Description

Updates VIP Real-Time Collaboration for Gutenberg 23.8’s unified real-time collaboration experiment.

- Enables the Gutenberg real-time collaboration experiment by default.
- Adds a VIP RTC settings control for disabling collaboration during an emergency.
- Disables the VIP provider and Gutenberg experiment when the setting is turned off.
- Hides the RTC control from Gutenberg’s Experiments page so it is managed only through the plugin.
- Preserves all unrelated Gutenberg experiment settings.
- Updates the minimum supported Gutenberg version to 23.8.0, and adds an admin notice for it.
- Updates the architecture documentation and links Gutenberg-owned behaviour to its upstream documentation.

## Manual testing

1. Use Gutenberg trunk in `.wp-env.json`. It'll need to be built. Alternately, Install and activate Gutenberg 23.8.0 or newer and the VIP Real-Time Collaboration plugin.
2. Go to **Settings → VIP Real-Time Collaboration**.
3. Confirm **Enabled (recommended)** is selected by default.
4. Open Gutenberg’s Experiments page and confirm the Real-Time Collaboration experiment is not shown.
5. Open the same post in two browser sessions using different users.
6. Confirm edits and collaborator presence are synchronized between both sessions.
7. Return to **Settings → VIP Real-Time Collaboration**, select **Disabled (emergency fallback)**, and save.
8. Confirm the disabled warning is displayed on the settings page.
9. Reload the editor and confirm real-time collaboration and the VIP WebSocket provider are no longer active.
10. Re-enable RTC from the plugin settings and save.
11. Reload the editor and confirm real-time collaboration works again.
12. If another Gutenberg experiment is enabled, toggle VIP RTC off and on and confirm the unrelated experiment remains enabled.

CleanShot 2026-08-19 at 08 10 27
CleanShot 2026-08-19 at 08 11 25
CleanShot 2026-08-25 at 09 00 09

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Please see [SECURITY.md](SECURITY.md).
 ### Prerequisites
 
 - WordPress 6.7+
-- [Gutenberg plugin](https://github.com/WordPress/gutenberg/releases) (v22.8.1 or newer)
+- [Gutenberg plugin](https://github.com/WordPress/gutenberg/releases) (v23.8.0 or newer)
   - The `wp-env`-based development environment automatically downloads Gutenberg from GitHub.
   - If you use a different development environment, ensure the Gutenberg plugin is installed and active.
 - Node.js
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -65,7 +65,7 @@ If you use a different development environment, ensure these variables are set a
 
 ### Yjs Inspector
 
-The development environment includes a Yjs Inspector tool for debugging collaborative document state. When you open a post in the editor, a link to the Yjs Inspector will be logged in your browser's development console. This link opens a web interface where you can inspect the current state ofthe Yjs document, view document history, and debug synchronization issues.
+The development environment includes a Yjs Inspector tool for debugging collaborative document state. When you open a post in the editor, a link to the Yjs Inspector will be logged in your browser's development console. This link opens a web interface where you can inspect the current state of the Yjs document, view document history, and debug synchronization issues.
 
 ## Guidelines
 
--- a/docs/SYSTEM_ARCHITECTURE.md
+++ b/docs/SYSTEM_ARCHITECTURE.md
@@ -17,17 +17,17 @@ The VIP Real-Time Collaboration plugin provides real-time collaborative editing 
 
 ### Key Technologies
 
-- **[Y.js (Yjs)](https://docs.yjs.dev/)**: [Conflict-free Replicated Data Types](https://crdt.tech/) (CRDTs) for operational transformation
+- **[Y.js (Yjs)](https://docs.yjs.dev/)**: [Conflict-free Replicated Data Types](https://crdt.tech/) (CRDTs) for convergent collaborative state
 - **[WebSockets](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API)**: Real-time bidirectional communication
 - **[JWT Authentication](https://jwt.io/introduction)**: Secure connection tokens
-- **WordPress Sync API**: Built on Gutenberg's experimental sync features
+- **WordPress Sync API**: Gutenberg's collaboration and CRDT infrastructure, exposed through `@wordpress/sync`
 
 ### High-Level Architecture
 
 ```mermaid
 graph TB
     A[WordPress Editor] --> B[Y.js CRDT Provider]
-    A --> C[Awareness Manager]
+    A --> C[Gutenberg Awareness]
     B --> D[WebSocket Server]
     C --> D
     D --> E[Other Connected Users]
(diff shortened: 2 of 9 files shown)


docs/CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ Please see [SECURITY.md](SECURITY.md).
 ### Prerequisites
 
 - WordPress 6.7+
-- [Gutenberg plugin](https://github.com/WordPress/gutenberg/releases) (v22.8.1 or newer)
+- [Gutenberg plugin](https://github.com/WordPress/gutenberg/releases) (v23.8.0 or newer)
```
ingeniumed, line 14: The minimum Gutenberg version is now 23.8.0 because that is the first release containing the unified Real-Time Collaboration experiment introduced in Gutenberg PR #80658. The plugin now relies exclusively on that experiment and no longer supports the previous RTC option model.

Given the direction we plan to go for RTC within Gutenberg, we shouldn't attempt backwards compatibility.

docs/SYSTEM_ARCHITECTURE.md
ingeniumed, line 1: The architecture document now links to Gutenberg for CRDT lifecycle, persistence, awareness, polling, and provider behaviour because Gutenberg owns those parts of the system. Keeping the plugin documentation focused on the VIP WebSocket integration avoids duplicating upstream implementation details that could become outdated.

pkevan (commented):
Two blockers for a default-on flip:

1. **No migration.** Trunk read `get_option( 'wp_collaboration_enabled' )`; this reads a new option and returns `true` when it's absent — which it always is on upgrade. Sites that deliberately disabled RTC get it re-enabled silently. Gutenberg handled exactly this when it renamed the same option, seeding the old value forward only when it was falsy.

2. **No programmatic kill switch.** Grepping every `apply_filters` in `inc/` turns up `vip_rtc_support_email` and two in `SyncPermissions` — nothing for enablement, no constant, nothing honouring `WP_ALLOW_COLLABORATION`. For a fleet-wide default-on, the only off-switch is a radio button per site.

One thing I couldn't verify: no local Gutenberg checkout has a `gutenberg-real-time-collaboration` experiment (newest here is 23.7.0-rc.1; all still gate on `wp_collaboration_enabled`). Since Gutenberg reads `gutenberg-experiments` at include time in `lib/load.php` and these filters register on `plugins_loaded`, worth confirming against real 23.8 source that the filter lands at all.

chriszarate:
> default-on flip

What is the context? Is this in reference to this PR or a predicted future one? In Gutenberg code or elsewhere?

shekharnwagh:
Locally, this works as expected. I am worried that it added another way to enable or disable RTC beyond toggling the integration. Besides being slightly faster, what does it add that the integration toggle can't handle?

ingeniumed:
> 1. **No migration.** Trunk read `get_option( 'wp_collaboration_enabled' )`; this reads a new option and returns `true` when it's absent — which it always is on upgrade. Sites that deliberately disabled RTC get it re-enabled silently. Gutenberg handled exactly this when it renamed the same option, seeding the old value forward only when it was falsy.

Gutenberg deletes the legacy option as part of its migration in making RTC an experiment. So this is not likely to work. The experiment is on by default for those who want to install this plugin as that's the behaviour we prefer. In the case where RTC isn't needed, the integration would be switched off or the plugin would be removed.
 
> 2. **No programmatic kill switch.** Grepping every `apply_filters` in `inc/` turns up `vip_rtc_support_email` and two in `SyncPermissions` — nothing for enablement, no constant, nothing honouring `WP_ALLOW_COLLABORATION`. For a fleet-wide default-on, the only off-switch is a radio button per site.

That constant has been removed as part of the move in making RTC an experiment. None of that was committed to core so it was removed. The experiment is the way to turn it on and off besides the `sync.providers` that guards any providers in active use.

> One thing I couldn't verify: no local Gutenberg checkout has a `gutenberg-real-time-collaboration` experiment (newest here is 23.7.0-rc.1; all still gate on `wp_collaboration_enabled`). Since Gutenberg reads `gutenberg-experiments` at include time in `lib/load.php` and these filters register on `plugins_loaded`, worth confirming against real 23.8 source that the filter lands at all.

This is [included in Gutenberg 23.8](https://github.com/WordPress/gutenberg/pull/80658#issuecomment-5337824110) which has been released and is testable using that.

Alternately, you can change `.wp-env.json` to use a local Gutenberg trunk checkout and that will have the changes.

ingeniumed:
> Locally, this works as expected. I am worried that it added another way to enable or disable RTC beyond toggling the integration. Besides being slightly faster, what does it add that the integration toggle can't handle?

I recall one of the issues being that not everyone has access to the integration toggle, and it's useful to have a "break glass in case of emergency" switch. The purpose for this toggle is for just that. 

If the editorial experience is blocked then it can be immediately fixed, instead of getting someone with the proper admin rights to unblock the team.

That said, I'm fine if we want to take it out for now.

@smithjw1 - what do you think?

pkevan:
> > default-on flip
> 
> What is the context? Is this in reference to this PR or a predicted future one? In Gutenberg code or elsewhere?

Who knows - this was an agent review that i didn't review.

One thought is that with the return to an experiment and the change of option names, at some point we might inadvertently reuse a previous name (the risk is low), should we keep a record of the historic permutations in tests to avoid?

ingeniumed:
> > > default-on flip
> > 
> > 
> > What is the context? Is this in reference to this PR or a predicted future one? In Gutenberg code or elsewhere?
> 
> Who knows - this was an agent review that i didn't review.
> 
> One thought is that with the return to an experiment and the change of option names, at some point we might inadvertently reuse a previous name (the risk is low), should we keep a record of the historic permutations in tests to avoid?

I mentioned in https://github.com/Automattic/vip-real-time-collaboration/pull/225#issuecomment-5388316162 that as part of the migration to Gutenberg 23.8 all previous options will be deleted. So that isn't too big of a worry IMO.

inc/Compatibility/Compatibility.php
```diff
@@ -32,9 +32,6 @@ public static function admin_notices(): void {
 
 	/**
 	 * Check if the Gutenberg plugin is active.
-	 *
-	 * TODO: Check GUTENBERG_VERSION in production to ensure it is running a
-	 * compatible version.
```
chriszarate, line 37: Can we do this? If it's not possible, can we explain why we can't do it in this docblock?

vip-real-time-collaboration.php
```diff
@@ -53,7 +53,7 @@
 
 	Settings::init();
 
-	// If RTC is disabled, return early.
+	// Keep the runtime gate independent from Gutenberg's filtered experiment value.
```
chriszarate: Not an improvement IMO. I'm not really sure what this means

chriszarate (approved):
Marked a few places where comments need improvement

vip-real-time-collaboration.php
ingeniumed: Done in https://github.com/Automattic/vip-real-time-collaboration/pull/225/commits/43c81067027d0db4eae94fc4f7f2dedb6b52985c

inc/Compatibility/Compatibility.php
ingeniumed: I've gone ahead and added the version check in https://github.com/Automattic/vip-real-time-collaboration/pull/225/commits/43c81067027d0db4eae94fc4f7f2dedb6b52985c

inc/Compatibility/Compatibility.php
```diff
@@ -32,14 +46,31 @@ public static function admin_notices(): void {
...
 		return defined( 'IS_GUTENBERG_PLUGIN' ) && constant( 'IS_GUTENBERG_PLUGIN' );
 	}
 
+	/**
+	 * Check if the active Gutenberg plugin version is compatible.
+	 */
+	private static function is_gutenberg_plugin_version_compatible(): bool {
+		if ( ! defined( 'GUTENBERG_VERSION' ) ) {
+			return false;
+		}
```
chriszarate, line 60: Is this compatible with Gutenberg development builds? Is there an escape hatch if someone is testing Gutenberg changes with this plugin?

ingeniumed: This is compatible with Gutenberg development builds since the version is always defined correctly. I confirmed it by testing it locally with:

- Local trunk build
- Latest release build
- Old release build

ingeniumed: The only change that'd be made is within `.wp-env.json` to use the local GB build.

ingeniumed: Discussed offline that this is good.

PR #226, opened 2026-08-24
Use Y shim instead of wp.sync bundle

alecgeatches:
Gutenberg is removing `wp.sync` and will instead pass the editor's Yjs module to sync providers as a `Y` option in https://github.com/WordPress/gutenberg/pull/81999. This PR updates the WebSocket transport to consume the new option. It stays backwards-compatible with Gutenberg versions that still expose `wp.sync.Y`.

## How the shim avoids a second Yjs instance

Our bundle includes `y-websocket` and `y-protocols`, which statically import `yjs`. If webpack resolved that import normally, the plugin would ship its own copy of Yjs alongside the editor's copy, and two Yjs instances working on the same document silently corrupt data (see [yjs#438](https://github.com/yjs/yjs/issues/438)).

Instead, the build now aliases the `yjs` module name to `src/yjs-shim.ts`, so those libraries load the shim instead of bundling a second Yjs. The shim re-exports the functions the libraries actually use, initially with no value. Because module exports are live bindings, the libraries read the current value at call time rather than at load time, and they only call Yjs functions after a provider is constructed. That makes it safe to fill in the shim when the provider creator runs. An inert initial Yjs import exists, but is replaced with the working singular Gutenberg `Y` export before any code is actually run.

## Testing

I updated `.wp-env.json`'s Gutenberg plugin to point to a local copy of https://github.com/WordPress/gutenberg/pull/81999, with WordPress 7.1 (which doesn't already import `wp.sync`):

```js
{
  /* ... */
  "core": "WordPress/WordPress#7.1",
  "plugins": [ ".", "/Users/alec/<path-to-gutenberg-pr-checkout>" ]
}
```

Additionally, run the WebSockets server in a terminal:

```bash
$ VIP_RTC_WS_AUTH_SECRET=vip_rtc_ws_auth_secret npm run dev:websocket-server
```

After everything is running, test RTC works without a `wp.sync` global:

https://github.com/user-attachments/assets/c68da4b4-5f07-4117-9948-39d885d746c7

### DIFF
--- a/inc/Assets/Assets.php
+++ b/inc/Assets/Assets.php
@@ -15,6 +15,7 @@ use function wp_add_inline_script;
 use function wp_create_nonce;
 use function wp_die;
 use function wp_enqueue_script;
+use function wp_script_is;
 
 /**
  * Enqueues the necessary JavaScript and CSS assets for the plugin.
@@ -54,7 +55,16 @@ final class Assets {
 		 */
 		$asset = include $asset_file;
 
-		$dependencies = array_unique( array_merge( $asset['dependencies'], [ 'wp-sync' ] ) );
+		$dependencies = $asset['dependencies'];
+
+		// Gutenberg versions before WordPress/gutenberg#81999 expose the editor's
+		// Yjs instance on the `wp.sync` global, which our bundle reads at load
+		// time. Newer versions no longer register the `wp-sync` script handle and
+		// pass Yjs to the provider creator instead, so only depend on the handle
+		// when it exists. An unregistered dependency would block the enqueue.
+		if ( wp_script_is( 'wp-sync', 'registered' ) ) {
+			$dependencies = array_unique( array_merge( $dependencies, [ 'wp-sync' ] ) );
+		}
 
 		wp_enqueue_script(
 			'vip-real-time-collaboration',
--- a/src/types/sync.ts
+++ b/src/types/sync.ts
@@ -1,7 +1,8 @@
 import type {
 	ConnectionStatus as WordPressConnectionStatus,
-	ProviderCreatorOptions,
+	ProviderCreatorOptions as WordPressProviderCreatorOptions,
 } from '@wordpress/sync';
+import type * as YjsModule from 'yjs';
 
 type ConnectedStatus = Exclude< WordPressConnectionStatus, { status: 'disconnected' } >;
 type DisconnectedStatus = Extract< WordPressConnectionStatus, { status: 'disconnected' } >;
@@ -45,8 +46,17 @@ export interface ProviderCreatorResult {
 	on: ProviderOn;
 }
 
+/**
+ * WordPress's provider creator options widened with the `Y` option added in
+ * WordPress/gutenberg#81999. Newer Gutenberg versions pass the editor's Yjs
+ * module here. Older versions expose it as the `wp.sync.Y` global instead, so
+ * the option stays optional. This local widening can be removed once the
+ * installed `@wordpress/sync` types include `Y`.
+ */
+export type ProviderCreatorOptions = WordPressProviderCreatorOptions & {
+	Y?: typeof YjsModule;
+};
+
 export type ProviderCreator = (
 	options: ProviderCreatorOptions
 ) => Promise< ProviderCreatorResult >;
-
-export type { ProviderCreatorOptions };
--- a/src/websocket-client.ts
+++ b/src/websocket-client.ts
@@ -26,6 +26,7 @@ import {
 } from '@/utilities/room-client-limit';
 import { SyncConnectionStatusEmitter } from '@/utilities/sync-event-emitter';
 import { getWebSocketClosePolicy, getWebSocketCloseScope } from '@/websocket-close-policy';
+import { setYjsModule } from '@/yjs-shim';
 
 import type {
 	ConnectionStatus,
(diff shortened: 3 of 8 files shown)


chriszarate:
Excited to see this approach working. We originally tried a similar approach and ran into race conditions where our code or library code needed access to `Y` before the provider had been created.

Perhaps the code has changed enough to eliminate these race conditions, but that surrounds my main feedback:

- Has the race condition been fully mitigated in this codebase?
- Has the race condition been fully mitigated in other codebases (if they exist)?
- Is this approach straightforward enough for other codebases to adopt (if they exist)?

ingeniumed:
Noting that with the merge of #225, we only support Gutenberg versions 23.8 and above. This should simplify backwards compatibility support.

chriszarate:
> This should simplify backwards compatibility support.

How so?

alecgeatches:
> Excited to see this approach working. We originally tried a similar approach and ran into race conditions where our code or library code needed access to `Y` before the provider had been created.

Thank you for taking a look and the feedback! See below:

---

> * Has the race condition been fully mitigated in this codebase?

I believe so. We had a different race condition before from when Yjs was read at load time. `externals: { yjs: 'wp.sync.Y' }` read the global while our bundle was evaluating and depended on `wp-sync` being available.

The shim is filled [at the top of the provider creator](https://github.com/Automattic/vip-real-time-collaboration/blob/cbf7589ee2a76f5ebb9918dd539afb2e2bfc4086/src/websocket-client.ts#L347-L351) now, before the provider is constructed. The provider is also created with `connect: false`, so the first real Yjs call (writing sync step 1) does not happen until after an auth token fetch succeeds. In our code, `Y` is filled with the live reference from Gutenberg a few steps before it's ever used.

Whether or not this is safe relies on if any bundled code uses a Yjs symbol/function before `setYjsModule()` runs. For our Yjs library usages now:

- y-websocket v3.0.0 [imports the `yjs` namespace](https://github.com/yjs/y-websocket/blob/v3.0.0/src/y-websocket.js#L7) but never dereferences it at runtime. Every `Y.` in the file is inside a JSDoc comment.
- y-protocols v1.0.7 has exactly three runtime uses, all inside function bodies: [`Y.encodeStateVector` in `writeSyncStep1`](https://github.com/yjs/y-protocols/blob/v1.0.7/sync.js#L50), [`Y.encodeStateAsUpdate` in `writeSyncStep2`](https://github.com/yjs/y-protocols/blob/v1.0.7/sync.js#L61), and [`Y.applyUpdate` in `readSyncStep2`](https://github.com/yjs/y-protocols/blob/v1.0.7/sync.js#L84). [awareness.js](https://github.com/yjs/y-protocols/blob/v1.0.7/awareness.js) and [auth.js](https://github.com/yjs/y-protocols/blob/v1.0.7/auth.js) have none.

Nothing reads Yjs during module evaluation. I don't see any stated guarantee of this, but it seems structured this way and holds up for the versions of Yjs we use.

---

> * Has the race condition been fully mitigated in other codebases (if they exist)?

This is somewhat impossible to say, but given that Yjs' internal libraries are set up to avoid `Y` symbol usages until run time, it doesn't seem likely. Some possibilities might be if a provider uses a global `Y.Doc` instantiated at the top level, or includes its own version of Yjs (already bad), but I'm doubting that existing implementations are wildly different from ours here.

Part of [the discussion on the Gutenberg side](https://github.com/WordPress/gutenberg/pull/81999#issuecomment-5413321013) is to ensure we reach out to known providers before we remove `wp.sync`, so we can ideally test this theory before making it required. A first step is this PR as a proof-of-concept.

---

> * Is this approach straightforward enough for other codebases to adopt (if they exist)?

  I think so. The entire adaptation here is [a short shim file](https://github.com/Automattic/vip-real-time-collaboration/blob/cbf7589ee2a76f5ebb9918dd539afb2e2bfc4086/src/yjs-shim.ts) plus [one webpack alias](https://github.com/Automattic/vip-real-time-collaboration/blob/cbf7589ee2a76f5ebb9918dd539afb2e2bfc4086/webpack.utils.js#L39-L45). Any modern bundler should support this.

Two things need to be true for someone to use this pattern. First, the libraries they bundle must only touch Yjs inside functions, which is true of `y-websocket` and `y-protocols` per the links above, and likely of transports generally since they operate on a doc that's handed to them. Second, the shim must export the symbols those libraries use. We only use a handful of Yjs functions directly:

https://github.com/Automattic/vip-real-time-collaboration/blob/cbf7589ee2a76f5ebb9918dd539afb2e2bfc4086/src/yjs-shim.ts#L33

but this would need to be widened to include more API calls if needed in the transport. Again, I'm not expecting there to be very complex transports in the wild, and expect they'll follow a similar setup path to ours here. We'll have to test that in practice.

chriszarate:
Thanks for the rundown. It's good to see that library code isn't touching `Y` during module evaluation. This is still gotcha-land, and I can invent a couple more ways this shim might fail:

- A destructuring import (`import { encodeStateAsUpdate} from 'yjs'`) might access `Y` too early.
- It's possible that some bundler might transform a `*` import into a destructuring one in pursuit of tree shaking.

That said, this is a breaking API change to an API that we said would break. The way forward is navigable. As long as we document it (on the Gutenberg side), this seems like a reasonable ask.

chriszarate (approved):

ingeniumed:
> > This should simplify backwards compatibility support.
> 
> How so?

I was referring to 

> It stays backwards-compatible with Gutenberg versions that still expose wp.sync.Y.

My original thought -> Since our plugin requires Gutenberg v23.8 as the minimum supported version this isn't a worry anymore. That's the version after which RTC became an experiment.

What I hadn't realized was that the other PR within GB hadn't been merged in yet, and so 23.8 didn't contain the resulting fix. I DM'd Alec this but didn't update my comment to reflect that.

PR #121, opened 2025-12-04
Add connection limit to WebSocket server

shekharnwagh:
# Add connection limit to WebSocket server

## Summary

Adds configurable connection limits to the WebSocket server to prevent overload and maintain stability. When the limit is reached, new connections are rejected with a clear error message.

## Changes

- **Connection limit enforcement**: Rejects connections when `MAX_CONNECTIONS` is reached
- **Configurable via environment**: Set `VIP_RTC_WS_MAX_CONNECTIONS` (default: 1500, or -1 for unlimited)
- **Structured error responses**: Returns HTTP 503 with JSON body containing error code and message
- **Metrics tracking**: New Prometheus counter `wpvip_rtc_connection_limit_reached_total`

## Testing

Test with websocat and curl to verify:
- Connections are properly rejected when limit is reached
- Error response includes correct status code (503) and JSON body
- Error code and message are user-friendly and actionable

### Websocat

```sh
websocat "ws://localhost:1234/site-1/postType/post-{postId}?auth={JWT_TOKEN}"
```

### Curl

Useful to inspect the HTTP response body message.

```sh
curl -s -i \
	-H "Connection: Upgrade" \
	-H "Upgrade: websocket" \
	-H "Sec-WebSocket-Version: 13" \
	-H "Sec-WebSocket-Key: {WEBSOCKET_KEY}" \
	"http://localhost:1234/site-1/postType/post-{postId}?auth={JWT_TOKEN}"
```

PR #237, opened 2026-09-17
Fix modal flicker during WebSocket retries

shekharnwagh:
## Description

During retries, `connecting` and early `connected` events replace the previous error before the connection recovers. This caused error modals to flicker or change their message.

This change preserves the last error and retry metadata throughout reconnect attempts and clears them only after the retry succeeds.


src/websocket-client.ts
```diff
@@ -475,12 +477,23 @@ export function createWebSocketConnection(
 				 * Skip 'disconnected' status - handled in connection-close above to preserve error details.
 				 * y-websocket emits 'connection-close' (with error code) then 'status: disconnected' (no error).
 				 */
-				if ( event.status !== 'disconnected' ) {
+				// Preserve the last failure and retry metadata throughout retries.
+				// Raw WebSocket open can precede rejection, so only successful
+				// connection clears the error.
```
chriszarate: This isn't really clear. Does this mean that the connection can close with an error after the `'connected'` event fires? Is that why you're using `'sync'` (which has nothing to do with network connection status)?

Can you make this commentary more clear?

shekharnwagh: >  Does this mean that the connection can close with an error after the 'connected' event fires? Is that why you're using 'sync' (which has nothing to do with network connection status)?

Yes, if we reject the upgrade request directly, the client can't read the error code/message ( see the earlier [thread](
https://github.com/Automattic/vip-real-time-collaboration/pull/121#discussion_r2590217651) on #121 ). So we upgrade the request and then close it immediately with an error code and message, which the client can read. So `sync` is the first event that fires after the WebSocket server has actually accepted the connection. Updated the comments in e7344e1

chriszarate (approved):

src/websocket-client.ts
chriszarate: Specifically `sync` fires after the client receives `SyncStep2` from the server (acknowledging their `SyncStep1` and delivering any missing updates)

